PR #9, opened 2020-06-24
Linter

ykjchen:
Copied https://github.com/brettallenyo/step2020/pull/1

To run locally:
```
make pretty
make validate
```

Also includes a workflow to run the above on code push.


ykjchen:
As discussed during sync, I reversed the commit that added the workflow. We can re-add that back in once existing validations have been resolved.

PR #14, opened 2020-06-26
Build out blog page with posts & side nav bar; Format code of index.html & gallery.html & add slight features (e.g. links)

annemariecabs:
Fix several CSS & HTML elements in gallery; Format blog & add posts; Add JS button & build out side nav in blog.html

Gallery Page: Update file names to match html so images will display. Format tags better.
About Me Page: Format page better with 'make pretty'. Add links to elements (e.g. Princeton University header links to Princeton's website).
Blog Page: Build out side navigation bar, including with a "Feeling Lucky?" button that uses JS to skip to the start of a random post. Add fake posts (using lorem ipsum text) & format them with proper alignment, color scheme, etc in CSS. 
pom.xml: Add project name in order to properly deploy.


agcaballero/portfolio/src/main/webapp/script.js
```diff
@@ -11,3 +11,46 @@
 // WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 // See the License for the specific language governing permissions and
 // limitations under the License.
+
+/* Returns the link to a random post on the blog */
```
ykjchen, line 15: JSDoc style prescribes `/** Returns the link to a random post on the blog */` (slash-star-star ...)
https://google.github.io/styleguide/jsguide.html#jsdoc-general-form

This comment also applies to functions below.

agcaballero/portfolio/src/main/webapp/script.js
```diff
@@ -11,3 +11,46 @@
...
+  const posts = ['#post1-lnk', '#post2-lnk', '#post3-lnk', '#post4-lnk'];
+
+  // returns a random link
+  return posts[Math.floor(Math.random() * posts.length)];
+}
+
+/* Tests randomPost() by verifying it spits out only one of the random post
+options & logging the percentage of times it returns each */
+function testRandomPost(trials = 10000) {
+  const LENGTH = 4;
```
ykjchen, line 27: I like how you tried to follow constant naming convention, but my feeling is that this doesn't count as a constant, since it's used only locally within a single function.

I can infer that LENGTH is the number of items in `posts` in the randomPost() function, but a more readable strategy, and one requiring less maintenance, would be to declare the posts separately, either as a constant or the return value of another function:

```
function postLinks() {
  return ['#post1-lnk', '#post2-lnk', '#post3-lnk', '#post4-lnk'];
}

function randomPost() {
  // array with all post links
  const posts = postLinks();

  // returns a random link
  return posts[Math.floor(Math.random() * posts.length)];
}

function testRandomPost(trials = 10000) {
  const postLinks = postLinks();
  const postLinkCount = postLinks.length;
  ...
}
```

This makes it clearer what `4` represents, and also avoid the need to update this value, if you add more post links.

Reference:
https://google.github.io/styleguide/jsguide.html#naming-constant-names

agcaballero/portfolio/src/main/webapp/script.js
```diff
@@ -11,3 +11,46 @@
...
+        break;
+      default:
+        console.log('Illegal response found.');
+        return;
+    }
+  }
+
+  /* Logs percentage of times each one occurred; should be ~ 25% */
+  for (let i = 0; i < LENGTH; i++) {
+    console.log(`${i}: ${sums[i] / trials * 100}%`);
```
ykjchen, line 52: Optional: consider logging the post link rather than the index, so that the reader of the log doesn't need to go through the code to figure out which post link corresponds to which index, i.e. instead of `${i}`, `${postLinks[i]}`, assuming you accepted the suggestion above to create a local `postLinks` const.

ykjchen, line 52: I like how you used a template literal here!

agcaballero/portfolio/src/main/webapp/script.js
```diff
@@ -11,3 +11,46 @@
...
+  }
+}
+
+testRandomPost(10000);  // runs 10000 trials on randomPost()
```
ykjchen, line 56: I wanted to point out that usually tests are separated from the tested functions. That said, I don't recommend spending time setting up a proper unit test suite here, and what you have here will suffice for this project.

I like how you thought about testing!

agcaballero/portfolio/src/main/webapp/script.js
```diff
@@ -12,19 +12,27 @@
...
-options & logging the percentage of times it returns each */
+/**
+Tests randomPost() by verifying it spits out only one of the random post
+options & logging the percentage of times it returns each
+*/
 function testRandomPost(trials = 10000) {
-  const LENGTH = 4;
+  const postLinks = postLinks();
+  const postLinkCount = postLinks.length;
   const sums = [0, 0, 0, 0];
```
ykjchen, line 36: Is there a way to make the number of elements in `sums` depend on `postLinkCount` too? For example, if another item was added to `postLinks`, this would need to have 5 elements.

annemariecabs: I think this would just be a declaration plus a simple for loop - I'll change it.

agcaballero/portfolio/src/main/webapp/script.js

agcaballero/portfolio/src/main/webapp/script.js

agcaballero/portfolio/src/main/webapp/script.js

agcaballero/portfolio/src/main/webapp/script.js

PR #31, opened 2020-07-01
7/1: Add footer & separate nav & footer into own html files 

annemariecabs:
Build HTML & CSS for a footer element. Create nav.html and footer.html for the navigation bar and footer html respectively. Add JS directly into HTML files to use jQuery to add the navigation bar and footer html into files (without needing the explicit HTML in all files using footer and/or navigation bar).

Update will allow for navigation bar and footer to be changed easily without needing to update every HTML file that represents a page the navigation bar and/or footer are on.


agcaballero/portfolio/src/main/webapp/style.css
```diff
@@ -229,3 +233,44 @@ nav button {
...
+
+#cols {
+  flex-direction: row;
+  justify-content: flex-end;
+}
+
+#cols .flex-col {
+  border-left: 1px solid white;
+  padding: 0px 10px;
+  margin: 5px;
```
ykjchen, line 266: alphabetical order puts margin above padding

annemariecabs: Done

PR #34, opened 2020-07-01
Week 3 Step Two - Uday

UdayKalra:
Implement JS button to fetch greeting from Servlet.


udaykalra/portfolio/src/main/java/com/google/sps/servlets/DataServlet.java
```diff
@@ -1,3 +1,4 @@
+
```
ykjchen, line 1: Revert extra newline at top of file?

udaykalra/portfolio/src/main/webapp/script.js
```diff
@@ -13,16 +13,10 @@
...
-  const greetings =
-      ['Hello world!', '¡Hola Mundo!', '你好，世界！', 'Bonjour le monde!'];
-
-  // Pick a random greeting.
-  const greeting = greetings[Math.floor(Math.random() * greetings.length)];
-
-  // Add it to the page.
-  const greetingContainer = document.getElementById('greeting-container');
-  greetingContainer.innerText = greeting;
+async function getDataUsingAsyncAwait() {
```
ykjchen, line 18: I'd recommend leaving "asyncAwait" out of the function name, since it's an implementation detail. In general, we strive to keep implementation details out of public API, since that makes it harder to change the API. For example, if the implementation is updated in the future to not use asyncAwait, then you'd need to change the function name, as well as references to the function.

Suggestion would be to name it like: `fetchCommentData()`, assuming that's what this is doing.

"Data" on its own is a bit generic, and could collide with other "get data" functions in the future.

udaykalra/portfolio/src/main/webapp/style.css
```diff
@@ -1,3 +1,8 @@
+#quote-container {
+  margin-top: 20px;
+  color: #fff;
```
ykjchen, line 3: Style encourages alphabetizing declarations.

udaykalra/portfolio/src/main/webapp/style.css
```diff
@@ -144,4 +149,5 @@ body {
   background-color: var(--body-color);
   height: 100%;
   margin: 0;
-}
\ No newline at end of file
+}
+>>>>>>> week-3-one
```
ykjchen, line 153: This looks like a conflict marker that needs to be removed.

UdayKalra: Done

udaykalra/portfolio/src/main/webapp/style.css
UdayKalra: Done

udaykalra/portfolio/src/main/webapp/script.js
UdayKalra: Done

udaykalra/portfolio/src/main/java/com/google/sps/servlets/DataServlet.java
UdayKalra: Done

.gitignore
```diff
@@ -0,0 +1,8 @@
+# Ignore generated Maven build files
+**/target/
+
+**/walkthroughs/
+
+package-lock.json
+node_modules/
```
ykjchen, line 7: Can you merge master into here? I think that should remove this file from the diff.

udaykalra/portfolio/src/main/webapp/script.js
```diff
@@ -13,16 +13,10 @@
 // limitations under the License.
 
 /**
- * Adds a random greeting to the page.
+ * Uses async and await to grab content from the data servlet.
```
ykjchen, line 16: I'd suggest leaving out implementation details from documentation too, as that increases maintenance cost of the documentation if the implementation changes. Instead, you can keep it generic:

```
Asynchronously fetches comment content from the data servlet.
```

udaykalra/portfolio/src/main/webapp/style.css
```diff
@@ -62,6 +62,11 @@
   --text-color: #e7e7e7;
 }
 
+#quote-container {
+  margin-top: 20px;
+  color: #fff;
```
ykjchen, line 67: Style prescribes alphabetical ordering of declarations.

.gitignore
UdayKalra: Will pull upstream after step 5 merge.

udaykalra/portfolio/src/main/webapp/style.css
UdayKalra: I had them alphabetical but separated by symbol, should they be alphabetical regardless of ./#/: ?

PR #45, opened 2020-07-06
6/30: Add home page with links to other pages on website

annemariecabs:
Add main home page (now in index.html), which will have a nav bar and display the different pages that the website has, move current "About me" code from the index.html file to about.html file.


agcaballero/portfolio/src/main/webapp/style.css
```diff
@@ -88,6 +96,75 @@ a:visited {
...
+.square {
+  align-content: center;
+  background-color: #e07a5f;
+  background-size: cover;
+  display: flex;
+  flex-direction: column;
+  height: 300px;
+  justify-content: center;
+  margin-top: 0px;
+  margin: 30px;
```
ykjchen, line 147: Does `margin` conflict with `margin-top`, since `margin: 30px` means 30px on all sides?

annemariecabs: I think margin-top overrides it. However, I can just set a more custom margin for this element to make it less confusing.

annemariecabs: Done

PR #46, opened 2020-07-06
Adds comments page with view limits and deletion

jelares:
Week three work for comments page, separate from main portfolio index page at the moment


jelares/portfolio/src/main/webapp/comments/commentFetcher.js
```diff
@@ -0,0 +1,48 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+/** Fetches comments and adds them to the DOM. */
+function getComments() {
+  const numCommentsElement = document.getElementById('num-comments');
+  const numComments = numCommentsElement.value;
+
+  /* This is the part about the query string which I mentioned in the sync */
```
jelares, line 20: @ykjchen This was what I was mentioning when I said I wasn't sure if I was implementing it correcty

jelares/portfolio/src/main/java/com/google/sps/servlet-data/ExampleComments.java
```diff
@@ -0,0 +1,63 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package com.google.sps.data;
+
+import java.util.ArrayList;
+import java.util.List;
+
+/** Class containing numComments example comments */
+public final class ExampleComments {
```
ykjchen, line 21: I like how you're thinking about encapsulation. Some thoughts:

Is `Example` meant to be in the class name? Does this represent comments that users may have written, or are these strictly "examples" that the website shows and are not editable?

If these represent all comments, including those added by users, consider removing "Example" from the name. Since this class represents a collection of x comments, consider naming it a "CommentBatch" or "CommentList". That said, it doesn't look like this class does more than an ArrayList, so you could consider just using an ArrayList instead, unless you have future plans for this class.

jelares/portfolio/src/main/java/com/google/sps/servlet-data/ExampleComments.java
```diff
@@ -0,0 +1,63 @@
...
+  private final List<String> comments;
+  private int numComments;
+
+  /**
+   * a mutable list of numComments comments
```
ykjchen, line 26: Consider using capitalization and punctuation, for easier reading, and as shown in examples:
https://google.github.io/styleguide/javaguide.html#s7-javadoc

jelares/portfolio/src/main/java/com/google/sps/servlet-data/ExampleComments.java
```diff
@@ -0,0 +1,63 @@
...
+  /**
+   * Add a new comment to the list
+   *
+   * @param comment the comment to add
+   */
+  public void addComment(String comment) {
+    comments.add(comment);
+    numComments++;
+  }
+}
```
ykjchen, line 63: A newline at the end removes this warning.

jelares/portfolio/src/main/java/com/google/sps/servlets/DataServlet.java
```diff
@@ -14,19 +14,99 @@
...
 import java.io.IOException;
+import java.util.ArrayList;
+import java.util.List;
 import javax.servlet.annotation.WebServlet;
 import javax.servlet.http.HttpServlet;
 import javax.servlet.http.HttpServletRequest;
 import javax.servlet.http.HttpServletResponse;
 
-/** Servlet that returns some example content. TODO: modify this file to handle comments data */
+/** Servlet that manages comments page */
```
ykjchen, line 33: Nit: "manages comments page" -> "manages the comments page."

jelares/portfolio/src/main/java/com/google/sps/servlets/DataServlet.java
```diff
@@ -14,19 +14,99 @@
...
+
+    // Query to find all comment entities sorted from newest to oldest
+    Query query = new Query("Comment").addSort("timestamp", SortDirection.ASCENDING);
+
+    DatastoreService datastore = DatastoreServiceFactory.getDatastoreService();
+    PreparedQuery results = datastore.prepare(query);
+
+    // adding all comment entities to an ExampleComments instance
+    final ExampleComments comments = new ExampleComments(new ArrayList<String>());
+    for (Entity entity : results.asIterable(FetchOptions.Builder.withLimit(commentLimit))) {
```
ykjchen, line 52: Do you know what's the difference between setting the limit here vs setting it on the Query?
https://cloud.google.com/datastore/docs/concepts/queries#datastore-datastore-limit-java

I'm wondering which is more applicable to this situation.

jelares/portfolio/src/main/java/com/google/sps/servlets/DataServlet.java
```diff
@@ -14,19 +14,99 @@
...
+      long id = entity.getKey().getId();
```
ykjchen, line 53: Is this used?

jelares/portfolio/src/main/java/com/google/sps/servlets/DataServlet.java
```diff
@@ -14,19 +14,99 @@
...
+       */
+      for (int commentNumber = 0; commentNumber < numberOfComments; commentNumber++) {
+        jsonComments += "\"comment" + commentNumber + "\": ";
+        jsonComments += "\"" + comments.getComment(commentNumber) + "\"";
+
+        if (commentNumber < (numberOfComments - 1)) {
+          jsonComments += ", ";
+        } else {
+          jsonComments += "}";
+        }
```
ykjchen, line 106: Is there a way to do this using an existing library like Gson? Advantages of using an existing lib:
1. already tested (you won't need to write tests to verify correctness of this),
2. reduces logic needed (ideally you'd be able to pass in a Map or key-value store to the library, and it would format it correctly for you).

jelares/portfolio/src/main/webapp/css/homeStyle.css
```diff
@@ -125,7 +125,7 @@ p {
 */
 
 #navbar {
-  padding: .5rem, 0rem, .5rem;
+  padding: 0.5rem, 0rem, 0.5rem;
```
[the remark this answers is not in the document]
jelares: Either make pretty or make validate actually did this, so maybe we should look into changing that

jelares/portfolio/src/main/webapp/comments/commentFetcher.js
jelares: There were no alternatives I had in mind in particular, if it looks good to you it works for me too

jelares/portfolio/src/main/java/com/google/sps/servlets/DeleteServlet.java
```diff
@@ -0,0 +1,49 @@
...
+    // Deletes all comment entities
+    for (Entity entity : results.asIterable()) {
+      Key commentKey = entity.getKey();
+      datastore.delete(commentKey);
+    }
+
+    // Redirect back to the HTML page.
+    response.sendRedirect("/comments/dataPage.html");
+  }
+}
```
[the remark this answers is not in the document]
jelares: Sure I can add newlines at the end, also what warning are you referring to?

jelares/portfolio/src/main/java/com/google/sps/servlet-data/ExampleComments.java
jelares: I could use an ArrayList however I thought that having an ADT could be useful in case I wanted to add more features in the future.

Also, you make a good point about the name - the idea was that the class was made for the purpose of having example comments for the walkthrough, but in reality the class does generally represent a batch of comments - I will rename it to something more appropriate.

jelares/portfolio/src/main/java/com/google/sps/servlets/DataServlet.java
jelares: In the walkthrough it said to use Gson or to write our own encoding, I wrote this one myself to practice encoding into JSON. When I go back in to address comments I will modify this part entirely to use Gson, I definitely agree with you in that using preexisting libraries is the best practice.

jelares/portfolio/src/main/java/com/google/sps/servlet-data/ExampleComments.java
jelares: To reiterate on this point, if time permits I plan on using this feature to store and retrieve blog-post entries on my portfolio from users. So in reality the ADT should represent any media object containing text field, author, title, date posted, etc. These are things which both comments and blog posts share, so I will rename the ExampleComments class to something along the lines of PostBatch

jelares/portfolio/src/main/java/com/google/sps/servlets/DataServlet.java
jelares: no, I removed it

jelares/portfolio/src/main/webapp/css/homeStyle.css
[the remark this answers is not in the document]
ykjchen: I see, it looks like prettier doesn't have a way to configure this:
https://github.com/prettier/prettier/issues/2705

Perhaps we can move to stylelint, but until then we can adopt a "project style" which allows leading zeros.

jelares/portfolio/src/main/java/com/google/sps/servlet-data/ExampleComments.java
ykjchen: An alternative would be to have a model object that represents a single comment or blog post, and let ArrayList handle collection-type things. For example:

```
class TextContent {
String text;
...
}

class Comment extends TextContent {
}

class BlogPost extends TextContent {
}
```

Storing and retrieving should probably be the responsibility of some other object(s) (see https://en.wikipedia.org/wiki/Single-responsibility_principle):

```
class CommentFetcher {
function ArrayList<Comment> fetchComments();
}
```

Or to generalize a bit you could have an entity fetcher with a parser object:

```
interface EntityParser {
function Comment parseEntity(Entity);
}

class CommentParser implements EntityParser {
}

class BlogPostParser implements EntityParser {
}

class EntityFetcher {
function Object fetch(String entityName, EntityParser parser);
}

```

jelares/portfolio/src/main/java/com/google/sps/servlets/DataServlet.java
ykjchen: AnneMarie pointed out that that API is unavailable to your environment. It does look like `withLimit()` is most applicable. Glad you found this!

jelares/portfolio/src/main/java/com/google/sps/servlets/DataServlet.java
ykjchen: I see, thanks for pointing out that part about the instructions.

I wonder if there's any higher-level API that could be used to make the formatting easier to read (even if less performant), especially at this point, without unit tests, the ability to easily manually inspect for errors is helpful.

Perhaps `String.join`?
https://mkyong.com/java/java-how-to-join-list-string-with-commas/

```
ArrayList<String> commentItems = new ...;
for (... commentNumber) {
  String commentItem = <e.g. 1:"comment text">;
  commentItems.add(commentItem);
}
return "{" + String.join(",", commentItems) + "}";
```

jelares: On second thought @ykjchen , I read through the gson documentation and didn't find any immediate way to serialize individual entries in an array with as separate attributes of a json object. Do you know any way I could use gson to replicate what my commentsToJson function does?

jelares/portfolio/src/main/java/com/google/sps/servlets/DataServlet.java
jelares: I have been trying to fix this for a few hours but cannot find any way to set the limit on Query itself. The link you sent me ended up being for the cloud datastore version, but for our project we are using the appengine datastore version - the documentation for it is here: https://cloud.google.com/appengine/docs/standard/java/javadoc/com/google/appengine/api/datastore/Query.html

However I have looked through it and cannot find any ways to set limit on Query objects. I'll keep looking tomorrow, though.

jelares/portfolio/src/main/java/com/google/sps/servlets/DataServlet.java
jelares: I used Gson with some custom code to solve this

jelares/portfolio/src/main/java/com/google/sps/servlet-data/ExampleComments.java
jelares: Updated the documentation

jelares/portfolio/src/main/java/com/google/sps/servlet-data/ExampleComments.java
jelares: Will keep this in mind for our future projects, as we talked about on gchat

jelares/portfolio/src/main/java/com/google/sps/servlets/DeleteServlet.java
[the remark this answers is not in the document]
jelares: Added newlines, thank you!

jelares/portfolio/src/main/webapp/comments/commentFetcher.js
```diff
@@ -0,0 +1,48 @@
...
+function deleteComments() {
+  /* Delete the comments then return the blank comments */
+  fetch('/delete-data', {
+    method: 'POST',
+  }).then(getComments);
+}
+
+/** Creates an <li> element containing text. */
+function createListElement(text) {
+  const liElement = document.createElement('li');
```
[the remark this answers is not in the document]
jelares: updated the names to not use abbreviations

jelares/portfolio/src/main/java/com/google/sps/servlet-data/PostBatch.java
```diff
@@ -0,0 +1,75 @@
...
+  }
+
+  /** Get the number of posts */
+  public int getNumPosts() {
+    return numPosts;
+  }
+
+  /** Get all posts as a string array */
+  public String[] getPostArray() {
+    String[] postArrayString = new String[numPosts];
```
ykjchen, line 47: Recommend naming this variable `posts`, since `postArrayString` sounds like it should be a `String` (maybe a string concatenated from the contents of array of posts).

jelares/portfolio/src/main/java/com/google/sps/servlet-data/PostBatch.java
```diff
@@ -0,0 +1,75 @@
...
+    this.numPosts = posts.size();
+  }
+
+  /** Get the number of posts */
+  public int getNumPosts() {
+    return numPosts;
+  }
+
+  /** Get all posts as a string array */
+  public String[] getPostArray() {
```
ykjchen, line 46: Is there a need to return String[] rather than List<String>? The latter return type would simplify implementation a lot, and also would likely be more useful for the consumer, since List has more API to offer.

```
public List<String> getPosts() {
  return posts;
}
```

jelares/portfolio/src/main/java/com/google/sps/servlets/DataServlet.java
ykjchen: Was the change pushed?

jelares/portfolio/src/main/webapp/comments/commentFetcher.js
```diff
@@ -0,0 +1,48 @@
...
+  const numCommentsElement = document.getElementById('num-comments');
+  const numComments = numCommentsElement.value;
+
+  /* This is the part about the query string which I mentioned in the sync */
+  fetch(`/data?num-comments=${numComments}`)
+      .then(response => response.json())
+      .then((comments) => {
+        const commentListElement = document.getElementById('comment-list');
+        commentListElement.innerHTML = '';
+        const commentText = Object.values(comments)[0];
```
ykjchen, line 26: It would be helpful to name this variable as a plural (e.g. commentStrings) to make it clearer that it's a collection.

Below the statement `commentText.length` sounds like it's retrieving the length of a string, rather than the length of a collection.

jelares/portfolio/src/main/java/com/google/sps/servlet-data/PostBatch.java
```diff
@@ -0,0 +1,75 @@
...
+  /**
+   * This ADT represents a mutable list of numPosts posts.
+   *
+   * Abstraction Function(posts) =  list of numPosts posts such that posts.get(i)
+   *                                represents post i where each post is a string of text.
+   * Representation Invariant: posts.size() == numPosts
+   * Saftey from Rep Exposure: all fields are never returned, numPosts only increased through
+   * mutator methods.
+   */
+
```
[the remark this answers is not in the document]
jelares: Sorry for the confusion, like I said in the other comment: this part of the documentation explains what the ADT (class itself) represents, the abstraction function, representation invariant, and safety from representation exposure argument.

jelares/portfolio/src/main/java/com/google/sps/servlet-data/PostBatch.java
jelares: I made a new getPostList() method in case the client would like a list version like you describe. The reason an array version exists is to use with gson, because the only way I could see to easily encode the comments with gson was with an array of strings.

jelares/portfolio/src/main/webapp/comments/commentFetcher.js
jelares: Thanks for the suggestion, did exactly that

jelares/portfolio/src/main/java/com/google/sps/servlet-data/PostBatch.java
jelares: I named it postsArray to match signify that is an array of posts, different from getPostList which returns a list of posts

jelares/portfolio/src/main/java/com/google/sps/servlet-data/PostBatch.java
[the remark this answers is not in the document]
jelares: I made a getPostList() method as well. I mentioned in another comment that while getPostArray is less useful than getPostList, it is still useful for ease of encoding with gson

jelares/portfolio/src/main/java/com/google/sps/servlet-data/PostBatch.java
```diff
@@ -0,0 +1,75 @@
...
+import java.util.ArrayList;
+import java.util.List;
+
+/** Class containing a numPosts-large batch of posts */
+public final class PostBatch {
+  private final List<String> posts;
+  private int numPosts;
+
+  /**
+   * This ADT represents a mutable list of numPosts posts.
```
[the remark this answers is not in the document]
jelares: Sorry if that wasn't clear - abstract data type. This part of the documentation explains what the ADT represents, the abstraction function, representation invariant, and safety from representation exposure argument.

jelares/portfolio/src/main/java/com/google/sps/servlet-data/PostBatch.java
```diff
@@ -0,0 +1,75 @@
...
+   */
+
+  public PostBatch(List<String> posts) {
+    this.posts = new ArrayList<>(posts);
+    this.numPosts = posts.size();
+  }
+
+  /** Get the number of posts */
+  public int getNumPosts() {
+    return numPosts;
```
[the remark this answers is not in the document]
jelares: you're right - that was left over from a previous iteration of the class

jelares/portfolio/src/main/java/com/google/sps/servlets/DataServlet.java
jelares: I had forgotten to remove the function but I did in the new commit

jelares/portfolio/src/main/java/com/google/sps/servlet-data/PostBatch.java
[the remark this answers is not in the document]
ykjchen: Oh interesting. It looks like Gson supports List too:
https://www.baeldung.com/gson-list

Or is there some other advantage you're thinking of?

jelares:
> Have you pushed the latest changes?

@ykjchen Yeah I pushed the new commit - does it not show up for you?

jelares/portfolio/src/main/java/com/google/sps/servlet-data/PostBatch.java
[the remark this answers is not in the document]
ykjchen: Is it possible to move the documentation next to the related members? That is, move the class documentation to the class Javadoc, and any method documentation to their respective methods? When I am trying to figure out how to use a class and it's methods, I tend to look directly at the class/method documentations, and am less likely to read through a class's full implementation to find other docs.

jelares/portfolio/src/main/java/com/google/sps/servlet-data/PostBatch.java
ykjchen: I think json.toJson() also supports List<String>?

ykjchen:
Changes showing up now, thanks.

jelares/portfolio/src/main/java/com/google/sps/servlets/DataServlet.java
```diff
@@ -27,43 +27,61 @@
...
+      */
+      for (int commentNumber = 0; commentNumber < numberOfComments; commentNumber++) {
+        jsonComments += "\"comment" + commentNumber + "\": ";
+        jsonComments += "\"" + comments.getComment(commentNumber) + "\"";
+
+        if (commentNumber < (numberOfComments - 1)) {
+          jsonComments += ", ";
+        } else {
+          jsonComments += "}";
+        }
```
aaronegolden: Can this manual JSON writing be handled by the gson JS-to-JSON library (I see you added this dependency in a later commit, can't tell immediately if that replaces this logic).

jelares: Yep, ended up using gson for it

PR #49, opened 2020-07-06
7/6: Step 5 of comments

annemariecabs:
Integrate comment function with datastore and add final styling for comments.


agcaballero/portfolio/src/main/java/com/google/sps/servlets/DataServlet.java
```diff
@@ -14,64 +14,82 @@
...
+    // Build a String of divs to hold comments to add to page
+    for (Entity entity : results.asIterable()) {
+      // Retrieve info from the Entity
+      Date timestamp = (Date) entity.getProperty("timestamp");
+      String name = (String) entity.getProperty("name");
+      String email = (String) entity.getProperty("email");
+      String comment = (String) entity.getProperty("comment");
+
+      // append current div to HTML string commentDivs
+      commentDivs.append(formatComment(timestamp, name, email, comment));
```
ykjchen, line 59: FYI: This is fine, though a possibly a more encapsulated approach would have been to create a constructor for a Comment that accepts an Entity. That is, to keep the Comment class, and create instances of Comments from instances of Entities.

agcaballero/portfolio/src/main/java/com/google/sps/servlets/DataServlet.java
```diff
@@ -14,64 +14,82 @@
...
+    // Respond to request with the commentDivs html
     response.setContentType("application/html;");
     response.getWriter().println(commentDivs);
   }
 
+  // Takes properties of a comment and formats them with proper HTML
+  private String formatComment(Date timestamp, String name, String email, String comment) {
+    // gets time in local time zone (default: US ET)
+    LocalDateTime ldt = new LocalDateTime(timestamp.getTime(), DateTimeZone.forID("US/Eastern"));
+    DateTimeFormatter fmt = DateTimeFormat.forPattern("h:mm a M/dd/yy");
```
ykjchen, line 71: In general non-standard abbreviations are discouraged, as they slow down reading. I'd suggest spelling these out instead (e.g. `dateTime` and `formatter`).

agcaballero/portfolio/src/main/java/com/google/sps/servlets/DataServlet.java
annemariecabs: I agree. It would be more modular, and also more understandable. I'll make the change.

agcaballero/portfolio/src/main/java/com/google/sps/servlets/DataServlet.java

agcaballero/portfolio/src/main/java/com/google/sps/servlets/DataServlet.java
```diff
@@ -47,31 +56,56 @@ public String toString() {
...
   }
 
-  // holds all comments that have been submitted in session so far
-  private final ArrayList<Comment> comments = new ArrayList<Comment>();
-
   @Override
   public void doGet(HttpServletRequest request, HttpServletResponse response) throws IOException {
+    // Retrieve from Datastore entities of type "Comment"
+    DatastoreService datastore = DatastoreServiceFactory.getDatastoreService();
+    Query query = new Query("Comment").addSort("timestamp", SortDirection.DESCENDING);
```
aaronegolden, line 90: I'm curious to know if the query or property fetch is case sensitive. I see capital "Comment" here but lowercase "comment" below.

annemariecabs: I think it must be. The "Comment" and "comment" actually refer to different things. "Comment" refers to the kind of entity, and "comment" is a property of this kind that actually holds the text. I think I'll change the variable name to "text" to make this more clear. Thank you!

annemariecabs: Done

PR #50, opened 2020-07-06
7/6: Step 6 of Comments Walkthrough (comment deletion & cap on viewable comments)

annemariecabs:
Add button and corresponding Java to delete all comments and add a input to change how many comments are shown on comments page at a time.


agcaballero/portfolio/src/main/java/com/google/sps/servlets/DataServlet.java
```diff
@@ -38,17 +38,25 @@
...
 public class DataServlet extends HttpServlet {
-
   @Override
   public void doGet(HttpServletRequest request, HttpServletResponse response) throws IOException {
+    // Retrieve max number of comments (if no info provided, default is 50)
+    int maxComments = Integer.parseInt(getParameter(request, "max-comments", "50"));
+
     // Retrieve from Datastore entities of type "Comment"
     DatastoreService datastore = DatastoreServiceFactory.getDatastoreService();
     Query query = new Query("Comment").addSort("timestamp", SortDirection.DESCENDING);
```
ykjchen, line 48: Can a limit be set so that we don't fetch more data than is needed?
https://cloud.google.com/datastore/docs/concepts/queries#cursors_limits_and_offsets

If so it seems more optimal to fetch only what's needed rather than fetch all and then discard the unneeded portion.

annemariecabs: The link is to the Cloud API for Datastore not the AppEngine API. As such, I can't apply that specific fix (to my knowledge) as it uses a different version of the Query class. I agree that the current code is not optimal, and I would prefer to fetch less data. I used the FetchOptions class to somewhat improve this, but I believe I'm still fetching unnecessary data. I can't seem to find a similar function/class that caps the number of entities a query fetches within the AppEngine API.

annemariecabs: It seems using FetchOptions does fix this issue, so this thread is resolved.

annemariecabs: Done

agcaballero/portfolio/src/main/java/com/google/sps/servlets/DeleteDataServlet.java
```diff
@@ -0,0 +1,48 @@
...
+
+    // Deltes comment entities
+    for (Entity entity : results.asIterable()) {
+      datastore.delete(entity.getKey());
+    }
+
+    // Redirect to the comments page
+    response.sendRedirect("comments.html");
+  }
+}
```
ykjchen, line 48: Newline at end resolves this warning.

agcaballero/portfolio/src/main/java/com/google/sps/servlets/DataServlet.java
```diff
@@ -41,14 +41,22 @@
...
     PreparedQuery results = datastore.prepare(query);
     StringBuilder commentDivs = new StringBuilder();
 
+    int count = 0;
+
     // Build a String of divs to hold comments to add to page
     for (Entity entity : results.asIterable()) {
+      // Check if max number of comments have been loaded
+      if(count == maxComments) break;
```
aaronegolden: Nit: newline before the break.

annemariecabs: I ended up removing this line to deploy a different fix. However, it's good to know that's stylistically better. Thank you!

agcaballero/portfolio/src/main/java/com/google/sps/servlets/DeleteDataServlet.java
annemariecabs: Done

PR #51, opened 2020-07-07
Edit files to smooth edges of merges from different branches 

annemariecabs:
Make sure the merges did not affect website. Edit nav bar to include all relevant pages. Add nav bar and footer to all page files. Fix style of files including CSS.

Note: the changes made can be seen as they have all been deployed on go/agcaballero-step-2020.

### DIFF
--- a/agcaballero/portfolio/src/main/webapp/comments.html
+++ b/agcaballero/portfolio/src/main/webapp/comments.html
@@ -5,18 +5,21 @@
     <title>Comments</title>
     <link rel="stylesheet" href="style.css" />
     <script src="script.js"></script>
+    <script
+      src="https://code.jquery.com/jquery-3.3.1.js"
+      integrity="sha256-2Kok7MbOyxpgUVvAk/HJ2jigOSYS2auK4Pfzbm7uH60="
+      crossorigin="anonymous"
+    ></script>
+    <script>
+      $(function () {
+        $("#nav").load("nav.html")
+        $("footer").load("footer.html")
+      })
+    </script>
   </head>
 
   <body onload="getComments()">
-    <div id="nav">
-      <ul id="nav-list" class="nav-ul">
-        <li class="nav-li"><a href="index.html">About</a></li>
-        <li class="nav-li"><a href="gallery.html">Gallery</a></li>
-        <li class="nav-li"><a href="blog.html">Blog</a></li>
-        <li class="nav-li"><a href="comments.html">Comments</a></li>
-        <li class="nav-li" id="last-elem">AnneMarie Caballero</li>
-      </ul>
-    </div>
+    <div id="nav"></div>
 
     <div class="content">
       <div class="header" id="comment-title">
@@ -68,5 +71,6 @@
         </div>
       </div>
     </div>
+    <footer></footer>
   </body>
 </html>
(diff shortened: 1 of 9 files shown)


agcaballero/portfolio/src/main/webapp/index.html
```diff
@@ -27,34 +27,49 @@ <h1 id="welcome-sign">Welcome!</h1>
...
-        <a href="contact.html"
-          ><div class="square" id="contact-square">
-            <h2>Contact Me</h2>
-            <hr />
-            <h3>If you have questions</h3>
-          </div></a
-        >
+        <div class="flex-container square-row">
+          <a href="about.html"
+            ><div class="square" id="abt-square">
```
ykjchen, line 32: Is it common to have the closing bracket on a separate line when it would fit on the same line as the opening bracket? I was expecting:

```
<a href="about.html">
  <div class="square" id="abt-square">
    ...
  </div>
</a>
```

annemariecabs: I don't think so-- I think it just got accidentally formatted strangely. Sometimes, I assume if slightly strange formatting happens, it was the result of make pretty, so changing it will go against the linter. However, I just checked and it doesn't seem to be a result of make pretty, so I'm changing it now.

annemariecabs:
@ykjchen I seem to be having issues pushing to master to update this PR, so I had to do it manually, and create a new PR. I'm linking to it here #53.

agcaballero/portfolio/src/main/webapp/index.html
```diff
@@ -27,34 +27,49 @@ <h1 id="welcome-sign">Welcome!</h1>
...
+            </div></a
+          >
+
+          <a href="contact.html"
+            ><div class="square" id="contact-square">
+              <h2>Contact Me</h2>
+              <hr />
+              <h3>If you have questions</h3>
+            </div></a
+          >
```
aaronegolden, line 71: Another case of weird newline-before-the-closing-brace here.

agcaballero/portfolio/src/main/webapp/comments.html
```diff
@@ -5,18 +5,21 @@
     <title>Comments</title>
     <link rel="stylesheet" href="style.css" />
     <script src="script.js"></script>
+    <script
+      src="https://code.jquery.com/jquery-3.3.1.js"
+      integrity="sha256-2Kok7MbOyxpgUVvAk/HJ2jigOSYS2auK4Pfzbm7uH60="
```
aaronegolden, line 10: Just for my own edification: What is the integrity hash used for?

agcaballero/portfolio/src/main/webapp/index.html
annemariecabs: Done

agcaballero/portfolio/src/main/webapp/comments.html
annemariecabs: I originally put it in there because the resource I learned how to use jquery from had the script tag with an integrity hash. I think [https://hacks.mozilla.org/2015/09/subresource-integrity-in-firefox-43/](url) this article explains it best

PR #54, opened 2020-07-08
Week 3 Step 6 - Uday

UdayKalra:
Add comment limiter and deletion features.

### DIFF
--- a/udaykalra/portfolio/src/main/java/com/google/sps/servlets/DataServlet.java
+++ b/udaykalra/portfolio/src/main/java/com/google/sps/servlets/DataServlet.java
@@ -59,7 +59,7 @@ public class DataServlet extends HttpServlet {
     ArrayList<String> comments = new ArrayList<String>();
 
     FetchOptions commentLimiter = FetchOptions.Builder.withLimit(limiter);
-    
+
     for (Entity entity : results.asIterable(commentLimiter)) {
       String commentText = (String) entity.getProperty("text");
       comments.add(commentText);
--- a/udaykalra/portfolio/src/main/java/com/google/sps/servlets/DataServlet.java
+++ b/udaykalra/portfolio/src/main/java/com/google/sps/servlets/DataServlet.java
@@ -55,10 +55,10 @@ public class DataServlet extends HttpServlet {
     Query query = new Query("CommentSingle").addSort("timestamp", SortDirection.DESCENDING);
 
     PreparedQuery results = datastore.prepare(query);
-    int limiter = getCommentCount(request);
+    int commentLimit = getCommentCount(request);
     ArrayList<String> comments = new ArrayList<String>();
 
-    FetchOptions commentLimiter = FetchOptions.Builder.withLimit(limiter);
+    FetchOptions commentLimiter = FetchOptions.Builder.withLimit(commentLimit);
 
     for (Entity entity : results.asIterable(commentLimiter)) {
       String commentText = (String) entity.getProperty("text");
@@ -90,22 +90,23 @@ public class DataServlet extends HttpServlet {
   }
 
   /**
-   * Gets count for comments to be shown.
+   * Gets count for comments to be shown. Only numbers 1 to 50 can be an input.
+   * Returns null on error.
    */
-  private int getCommentCount(HttpServletRequest request) {
+  private Integer getCommentCount(HttpServletRequest request) {
     // Get the input from the form.
     String commentCountString = request.getParameter("comment-count");
 
     // Convert the input to an int.
-    int commentChoice;
+    Integer commentCount;
     try {
-      commentChoice = Integer.parseInt(commentCountString);
+      commentCount = Integer.parseInt(commentCountString);
     } catch (NumberFormatException e) {
       System.err.println("Could not convert to int: " + commentCountString);
-      return -1;
+      return null;
     }
 
-    return commentChoice;
+    return commentCount;
   }
 
   /**
--- a/udaykalra/portfolio/src/main/java/com/google/sps/servlets/DataServlet.java
+++ b/udaykalra/portfolio/src/main/java/com/google/sps/servlets/DataServlet.java
@@ -17,6 +17,8 @@ package com.google.sps.servlets;
 import com.google.appengine.api.datastore.DatastoreService;
 import com.google.appengine.api.datastore.DatastoreServiceFactory;
 import com.google.appengine.api.datastore.Entity;
+import com.google.appengine.api.datastore.FetchOptions;
+import com.google.appengine.api.datastore.FetchOptions.Builder;
 import com.google.appengine.api.datastore.PreparedQuery;
 import com.google.appengine.api.datastore.Query;
 import com.google.appengine.api.datastore.Query.SortDirection;
@@ -56,12 +58,11 @@ public class DataServlet extends HttpServlet {
     int limiter = getCommentCount(request);
     ArrayList<String> comments = new ArrayList<String>();
 
-    int commentsGot = 1;
-    for (Entity entity : results.asIterable()) {
+    FetchOptions commentLimiter = FetchOptions.Builder.withLimit(limiter);
+    
+    for (Entity entity : results.asIterable(commentLimiter)) {
       String commentText = (String) entity.getProperty("text");
       comments.add(commentText);
-      if (commentsGot++ >= limiter)
-        break;
     }
 
     String outputJson = listToJson(comments);
--- a/udaykalra/portfolio/src/main/java/com/google/sps/servlets/DataServlet.java
+++ b/udaykalra/portfolio/src/main/java/com/google/sps/servlets/DataServlet.java
@@ -38,7 +38,7 @@ public class DataServlet extends HttpServlet {
    * Converts a ServerStats instance into a JSON string using the Gson library. Note: We first added
    * the Gson library dependency to pom.xml.
    */
-  private String ListToJson(List<String> inputList) {
+  private String listToJson(List<String> inputList) {
     Gson gson = new Gson();
     String json = gson.toJson(inputList);
     return json;
@@ -53,14 +53,18 @@ public class DataServlet extends HttpServlet {
     Query query = new Query("CommentSingle").addSort("timestamp", SortDirection.DESCENDING);
 
     PreparedQuery results = datastore.prepare(query);
-
+    int limiter = getCommentCount(request);
     ArrayList<String> comments = new ArrayList<String>();
+
+    int commentsGot = 1;
     for (Entity entity : results.asIterable()) {
       String commentText = (String) entity.getProperty("text");
       comments.add(commentText);
+      if (commentsGot++ >= limiter)
+        break;
     }
 
-    String outputJson = ListToJson(comments);
+    String outputJson = listToJson(comments);
 
     response.getWriter().println(outputJson);
   }
@@ -85,10 +89,29 @@ public class DataServlet extends HttpServlet {
   }
 
   /**
+   * Gets count for comments to be shown.
+   */
+  private int getCommentCount(HttpServletRequest request) {
+    // Get the input from the form.
+    String commentCountString = request.getParameter("comment-count");
+
+    // Convert the input to an int.
+    int commentChoice;
+    try {
+      commentChoice = Integer.parseInt(commentCountString);
+    } catch (NumberFormatException e) {
+      System.err.println("Could not convert to int: " + commentCountString);
+      return -1;
+    }
+
+    return commentChoice;
+  }
+
+  /**
    * Obtains parameter from Comments typing field.
    */
-
-  private String getRequestParameter(HttpServletRequest request, String comment, String defaultValue) {
+  private String getRequestParameter(
+      HttpServletRequest request, String comment, String defaultValue) {
     String value = request.getParameter(comment);
 
     if (value == null) {
--- a/udaykalra/portfolio/src/main/java/com/google/sps/servlets/DataServlet.java
+++ b/udaykalra/portfolio/src/main/java/com/google/sps/servlets/DataServlet.java
@@ -90,7 +90,8 @@ public class DataServlet extends HttpServlet {
   }
 
   /**
-   * Gets count for comments to be shown. Only numbers 1 to 50 can be an input.
+   * Gets count for comments to be shown.
+   * Asks for 1 comment if receives non-positive input.
    * Returns null on error.
    */
   private Integer getCommentCount(HttpServletRequest request) {
@@ -101,12 +102,15 @@ public class DataServlet extends HttpServlet {
     Integer commentCount;
     try {
       commentCount = Integer.parseInt(commentCountString);
+
     } catch (NumberFormatException e) {
       System.err.println("Could not convert to int: " + commentCountString);
       return null;
     }
-
-    return commentCount;
+    if (commentCount <= 0)
+      return 1;
+    else
+      return commentCount;
   }
 
   /**
--- a/udaykalra/portfolio/src/main/java/com/google/sps/servlets/DeleteDataServlet.java
+++ b/udaykalra/portfolio/src/main/java/com/google/sps/servlets/DeleteDataServlet.java
@@ -1,0 +1,54 @@
+// Copyright 2019 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     https://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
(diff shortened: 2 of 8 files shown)


udaykalra/portfolio/src/main/java/com/google/sps/servlets/DataServlet.java
```diff
@@ -53,14 +55,17 @@ public void doGet(HttpServletRequest request, HttpServletResponse response) thro
     Query query = new Query("CommentSingle").addSort("timestamp", SortDirection.DESCENDING);
 
     PreparedQuery results = datastore.prepare(query);
-
+    int limiter = getCommentCount(request);
```
ykjchen, line 58: I'd suggest naming this `commentLimit`, since `limiter` sounds like an object that does the limiting.

udaykalra/portfolio/src/main/java/com/google/sps/servlets/DataServlet.java
```diff
@@ -85,10 +90,29 @@ public void doPost(HttpServletRequest request, HttpServletResponse response) thr
   }
 
   /**
-   * Obtains parameter from Comments typing field.
+   * Gets count for comments to be shown.
```
ykjchen, line 93: Would be helpful to point out that -1 is returned if an error occurs.

It would be better practice to then define a file-scoped constant that represents "no comment limit", instead of relying on users of this method to use the literal -1. This would be less error-prone and more resilient to refactoring, as well as easier to read.

Alternatively, and maybe this is the preferable approach, this method could return Integer instead of int, where nil is returned if there is no limit. Then you wouldn't need to define a special value.

udaykalra/portfolio/src/main/java/com/google/sps/servlets/DataServlet.java
```diff
@@ -85,10 +90,29 @@ public void doPost(HttpServletRequest request, HttpServletResponse response) thr
...
    */
+  private int getCommentCount(HttpServletRequest request) {
+    // Get the input from the form.
+    String commentCountString = request.getParameter("comment-count");
+
+    // Convert the input to an int.
+    int commentChoice;
```
ykjchen, line 100: `commentChoice` sounds like it could be a string (it's the chosen comment).

I'd suggest `commentCount` instead, as that's more clearly a number, and also better shows that this number represents a limit.

udaykalra/portfolio/src/main/java/com/google/sps/servlets/DataServlet.java
```diff
@@ -85,10 +90,29 @@ public void doPost(HttpServletRequest request, HttpServletResponse response) thr
...
+    try {
+      commentChoice = Integer.parseInt(commentCountString);
+    } catch (NumberFormatException e) {
+      System.err.println("Could not convert to int: " + commentCountString);
+      return -1;
+    }
 
-  private String getRequestParameter(HttpServletRequest request, String comment, String defaultValue) {
+    return commentChoice;
```
ykjchen, line 108: What should be returned if it's 0. Is that a reasonable return value?

udaykalra/portfolio/src/main/java/com/google/sps/servlets/DeleteDataServlet.java
```diff
@@ -0,0 +1,53 @@
...
+public class DeleteDataServlet extends HttpServlet {
+  DatastoreService datastore = DatastoreServiceFactory.getDatastoreService();
+
+  /**
+   * Deletes Comment Data
+   */
+  @Override
+  public void doPost(HttpServletRequest request, HttpServletResponse response) throws IOException {
+    // Get all comments.
+    Query query = new Query("CommentSingle").addSort("timestamp", SortDirection.DESCENDING);
```
ykjchen, line 45: Probably don't need to sort if they're just going to be deleted? Not sorting avoids unnecessary work (could cost extra money!).

udaykalra/portfolio/src/main/webapp/script.js
```diff
@@ -22,10 +30,10 @@ function createCommentList(inputText) {
...
  * Fetches comments for display.
  */
-function getCommentData() {
-  fetch('/data')                          // sends a request to /data
-      .then(response => response.json())  // parses the response as JSON
-      .then((myComments) => {  // now we can reference the fields in myObject!
+function getCommentData(commentsLimit = 20) {
+  fetch('/data?comment-count=' + commentsLimit)  // sends a request to /data
+      .then(response => response.json())         // parses the response as JSON
+      .then((myComments) => {  // now we can reference the fields as an object!
```
ykjchen, line 36: Suggest renaming `myComments` to `comments`, since "my" is ambiguous here (who does it represent?).

udaykalra/portfolio/src/main/webapp/style.css
```diff
@@ -1,5 +1,6 @@
 .bottomRow {
   padding-top: 30px;
+  padding-bottom: 30px;
```
ykjchen, line 3: Alphabetical ordering...

Actually, I think there's a shorthand property, which is preferred by style.

udaykalra/portfolio/src/main/java/com/google/sps/servlets/DataServlet.java
```diff
@@ -85,10 +89,29 @@ public void doPost(HttpServletRequest request, HttpServletResponse response) thr
...
+  private int getCommentCount(HttpServletRequest request) {
+    // Get the input from the form.
+    String commentCountString = request.getParameter("comment-count");
+
+    // Convert the input to an int.
+    int commentChoice;
+    try {
+      commentChoice = Integer.parseInt(commentCountString);
+    } catch (NumberFormatException e) {
+      System.err.println("Could not convert to int: " + commentCountString);
```
aaronegolden, line 104: Good job validating this input/converting to an int. When I noticed you added a comment-count query parameter I was worried the (string) parameter might be put directly into a DB query, which would create SQL-injection vulnerabilities.

udaykalra/portfolio/src/main/java/com/google/sps/servlets/DataServlet.java
```diff
@@ -53,14 +53,18 @@ public void doGet(HttpServletRequest request, HttpServletResponse response) thro
     Query query = new Query("CommentSingle").addSort("timestamp", SortDirection.DESCENDING);
 
     PreparedQuery results = datastore.prepare(query);
-
+    int limiter = getCommentCount(request);
```
aaronegolden, line 58: I would suggest getting the limit before constructing the query, and then using it in a LIMIT clause in the query. SQL can take care of limiting results for you so you don't need to manually limit your output below.

udaykalra/portfolio/src/main/webapp/Contact.html
```diff
@@ -9,10 +9,7 @@
       integrity="sha384-ggOyR0iXCbMQv3Xipma34MD+dH/1fQ784/j6cY/iJTQUOhcWr7x9JvoRxT2MZw1T"
       crossorigin="anonymous"
     />
-    <link
-      rel="stylesheet"
-      href="https://cdnjs.cloudflare.com/ajax/libs/font-awesome/4.7.0/css/font-awesome.min.css"
-    />
+    <link rel="stylesheet" href="https://cdnjs.cloudflare.com/ajax/libs/font-awesome/4.7.0/css/font-awesome.min.css" />
```
aaronegolden, line 12: What is the licensing situation with this asset, and what is it used for (can we avoid it)?

UdayKalra: That's an import of Bootstrap. The license for it is here which is pretty freeing: https://getbootstrap.com/docs/4.0/about/license/ . I may be wrong but I believe the others' sites also use bootstrap.

udaykalra/portfolio/src/main/java/com/google/sps/servlets/DataServlet.java
UdayKalra: Done

udaykalra/portfolio/src/main/java/com/google/sps/servlets/DataServlet.java
UdayKalra: Changed to return null through Integer class.

udaykalra/portfolio/src/main/java/com/google/sps/servlets/DataServlet.java
UdayKalra: Done

udaykalra/portfolio/src/main/java/com/google/sps/servlets/DataServlet.java
UdayKalra: The site will only let numbers 1 through 50 go through, error checker is there for safe practice.

udaykalra/portfolio/src/main/java/com/google/sps/servlets/DeleteDataServlet.java
UdayKalra: Done

udaykalra/portfolio/src/main/webapp/script.js
UdayKalra: Done

udaykalra/portfolio/src/main/webapp/style.css
UdayKalra: Done

udaykalra/portfolio/src/main/java/com/google/sps/servlets/DataServlet.java
UdayKalra: I'm only seeing it possible online to limit a sorted query after the initial query through FetchOptions as is being done.

udaykalra/portfolio/src/main/webapp/Contact.html
UdayKalra: Correction: That's a deprecated image library and has been removed in next branch. Bootstrap is a different import.

udaykalra/portfolio/src/main/webapp/script.js
```diff
@@ -22,15 +30,18 @@ function createCommentList(inputText) {
...
+  fetch('/data?comment-count=' + commentsLimit)  // sends a request to /data
+      .then(response => response.json())         // parses the response as JSON
+      .then((commentsPackage) => {  // now we can reference the fields as an
+                                    // object!
         const commentsElement = document.getElementById('quote-container');
         commentsElement.innerHTML = '';
-        for (var increment = 0; increment < myComments.length; increment += 1) {
-          console.log(myComments);
-          commentsElement.appendChild(createCommentList(myComments[increment]));
+        for (var increment = 0; increment < commentsPackage.length;
```
ykjchen, line 40: `index` would be clearer than `increment`, since `increment` means some delta.

udaykalra/portfolio/src/main/java/com/google/sps/servlets/DataServlet.java
```diff
@@ -85,10 +90,30 @@ public void doPost(HttpServletRequest request, HttpServletResponse response) thr
...
+    Integer commentCount;
+    try {
+      commentCount = Integer.parseInt(commentCountString);
+    } catch (NumberFormatException e) {
+      System.err.println("Could not convert to int: " + commentCountString);
+      return null;
+    }
 
-  private String getRequestParameter(HttpServletRequest request, String comment, String defaultValue) {
+    return commentCount;
```
ykjchen, line 109: I see that the comment count is limited to >0 in other parts of the code, but functions are typically not expected to be aware of constraints made in other parts of a codebase (it would be too complex to keep track), and functions should be self-contained. For example, this function could in the future be called through a different code path, or the limits set in other parts of code may change.

So when defining API it helps to consider the full range of inputs; in this case, what you'd like to do if commentCount is 0 or negative.

udaykalra/portfolio/src/main/webapp/style.css
```diff
@@ -127,26 +123,30 @@ div.grey {
...
 
 html,
 body {
   background-color: var(--body-color);
-  height: 100%;
+  margin-bottom: 60px;
   margin: 0;
+  display: flex;
+  min-height: 100vh;
+  flex-direction: column;
```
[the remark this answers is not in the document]
UdayKalra: Will fix in next PR

udaykalra/portfolio/src/main/webapp/style.css
```diff
@@ -127,26 +123,30 @@ div.grey {
 div.main-content {
   background-color: var(--body-color);
   color: var(--main-color);
-  font: "Roboto", sans-serif;
-  height: 100%;
+  font: 'Roboto', sans-serif;
+
+  flex: 1;
```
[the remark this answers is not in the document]
UdayKalra: Will fix in next PR(merge conflict avoidance)

udaykalra/portfolio/src/main/java/com/google/sps/servlets/DataServlet.java
UdayKalra: Done

udaykalra/portfolio/src/main/webapp/script.js
UdayKalra: Done

udaykalra/portfolio/src/main/java/com/google/sps/servlets/DataServlet.java
UdayKalra: Set to return 1 comment.

udaykalra/portfolio/src/main/webapp/script.js
```diff
@@ -22,15 +30,18 @@ function createCommentList(inputText) {
...
  * Fetches comments for display.
  */
-function getCommentData() {
-  fetch('/data')                          // sends a request to /data
-      .then(response => response.json())  // parses the response as JSON
-      .then((myComments) => {  // now we can reference the fields in myObject!
+function getCommentData(commentsLimit = 20) {
+  fetch('/data?comment-count=' + commentsLimit)  // sends a request to /data
+      .then(response => response.json())         // parses the response as JSON
+      .then((commentsPackage) => {  // now we can reference the fields as an
```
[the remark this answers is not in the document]
UdayKalra: Done

udaykalra/portfolio/src/main/webapp/script.js
```diff
@@ -22,15 +30,18 @@ function createCommentList(inputText) {
...
         const commentsElement = document.getElementById('quote-container');
         commentsElement.innerHTML = '';
-        for (var increment = 0; increment < myComments.length; increment += 1) {
-          console.log(myComments);
-          commentsElement.appendChild(createCommentList(myComments[increment]));
+        for (var increment = 0; increment < commentsPackage.length;
+             increment += 1) {
+          console.log(commentsPackage);
+          commentsElement.appendChild(
+              createCommentList(commentsPackage[increment]));
```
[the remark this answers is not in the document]
UdayKalra: Done

udaykalra/portfolio/src/main/java/com/google/sps/servlets/DataServlet.java
```diff
@@ -85,10 +90,34 @@ public void doPost(HttpServletRequest request, HttpServletResponse response) thr
...
 
-  private String getRequestParameter(HttpServletRequest request, String comment, String defaultValue) {
+    } catch (NumberFormatException e) {
+      System.err.println("Could not convert to int: " + commentCountString);
+      return null;
+    }
+    if (commentCount <= 0)
+      return 1;
+    else
+      return commentCount;
```
ykjchen, line 113: Style requires braces, even when optional:
https://google.github.io/styleguide/javaguide.html#s4.1.1-braces-always-used

udaykalra/portfolio/src/main/java/com/google/sps/servlets/DataServlet.java
```diff
@@ -85,10 +90,34 @@ public void doPost(HttpServletRequest request, HttpServletResponse response) thr
   }
 
   /**
-   * Obtains parameter from Comments typing field.
+   * Gets count for comments to be shown.
+   * Asks for 1 comment if receives non-positive input.
```
ykjchen, line 94: Instead of "asks" (it's not clear what's being asked), clearer would be "returns":

```
Returns 1 if the comment limit specified in the request is 0 or negative.
```

udaykalra/portfolio/src/main/java/com/google/sps/servlets/DataServlet.java
UdayKalra: Will be changed in APIs PR

udaykalra/portfolio/src/main/java/com/google/sps/servlets/DataServlet.java
UdayKalra: Will be changed in APIs PR

udaykalra/portfolio/src/main/java/com/google/sps/servlets/DataServlet.java
UdayKalra: Changed in APIs PR

udaykalra/portfolio/src/main/java/com/google/sps/servlets/DataServlet.java
UdayKalra: Changed in APIs PR

PR #55, opened 2020-07-08
7/8: Integrate Maps API into Gallery page of website

annemariecabs:
Integrate Maps API; add map to Gallery page with markers on all the places photos were taken, rename some image files, hard-code the coordinates and other values for the photos, add infowindows to markers to display image and place when clicked on. 

Below is an image of what it looks like (as it is not deployed yet).
Screenshot 2020-07-08 at 7 20 01 PM


agcaballero/portfolio/src/main/webapp/script.js
```diff
@@ -72,6 +72,129 @@ function testRandomPost(trials = 10000) {
 
 testRandomPost(10000);  // runs 10000 trials on randomPost()
 
+/** Creates a Map using Map API */
+function createMap() {
+  const europeLatLng =
+      new google.maps.LatLng({lat: 48.499998, lng: 23.3833318});
+  // the locations of the photos
```
ykjchen, line 79: Maybe comment on why these are not stored in a database? For example:

```
These are hardcoded due to time constraints. Long term, these should be stored in a database.
```

agcaballero/portfolio/src/main/webapp/script.js
```diff
@@ -72,6 +72,129 @@ function testRandomPost(trials = 10000) {
...
+      '</div>';
+
+  const infoWindow = new google.maps.InfoWindow({content: content});
+  marker.addListener('click', () => {
+    infoWindow.open(map, marker);
+  });
+}
+
+// call for createMap() because it's only called within HTML
+createMap();
```
ykjchen, line 196: Is this needed just to satisfy validate?

agcaballero/portfolio/src/main/webapp/script.js
```diff
@@ -72,6 +72,129 @@ function testRandomPost(trials = 10000) {
...
+  const map = new google.maps.Map(
+      document.getElementById('map'), {center: europeLatLng, zoom: 4});
+
+  for (let i = 0; i < places.length; i++) {
+    const place = places[i];
+    createMarkerForDisplay(map, place.title, place.img, place.lat, place.lng);
+  }
+}
+
+/** Creates a marker that shows a read-only info window when clicked. */
```
ykjchen, line 176: What kind of info is displayed?

agcaballero/portfolio/src/main/webapp/script.js
```diff
@@ -72,6 +72,129 @@ function testRandomPost(trials = 10000) {
...
+function createMarkerForDisplay(map, title, img, lat, lng) {
```
ykjchen, line 177: Based on the documentation, I'd suggest naming this `displayImageInfoMarker` or something that captures the side effect of displaying the info. Methods named `create...` suggest that something is created and returned, without other side effects.

agcaballero/portfolio/src/main/webapp/script.js
```diff
@@ -72,6 +72,129 @@ function testRandomPost(trials = 10000) {
 
 testRandomPost(10000);  // runs 10000 trials on randomPost()
 
+/** Creates a Map using Map API */
```
ykjchen, line 75: Some more detail may be helpful, for example, letting the reader know that the map contains preset markers showing photos taken at different locations.

agcaballero/portfolio/src/main/webapp/script.js
```diff
@@ -72,6 +72,129 @@ function testRandomPost(trials = 10000) {
...
+function createMap() {
```
ykjchen, line 76: Nit: consider being more specific in the name, since this script file has a broad scope, for example, `createPhotoLocationMap`.

agcaballero/portfolio/src/main/webapp/script.js
annemariecabs: Thank you. I will add that. Currently, it's not stored in a database just because I was going to work on the server-side aspect of it today. However, I will add that comment as a place-holder.

agcaballero/portfolio/src/main/webapp/script.js
annemariecabs: @ykjchen I was unsure as to what the best practices would be for this. The method is called within the HTML, but validate still does not recognize that as the method being called. I would be happy to test this method, but I don't know how I would because it's mainly visual. I've had this problem with a couple of other methods as well in script.js.

agcaballero/portfolio/src/main/webapp/script.js
annemariecabs: Additionally, I was wondering if it would be better to store these in a CSV or in datastore.

agcaballero/portfolio/src/main/webapp/script.js
annemariecabs: That makes sense. Thank you!

agcaballero/portfolio/src/main/webapp/script.js
annemariecabs: It displays the title of the place and the image taken there. I will update the description to include that.

agcaballero/portfolio/src/main/webapp/script.js
annemariecabs: That makes sense. I agree.

agcaballero/portfolio/src/main/webapp/script.js
annemariecabs: I added more details. Thank you!

agcaballero/portfolio/src/main/webapp/script.js
annemariecabs: Done

agcaballero/portfolio/src/main/webapp/script.js

agcaballero/portfolio/src/main/webapp/script.js

agcaballero/portfolio/src/main/webapp/script.js

agcaballero/portfolio/src/main/webapp/script.js

agcaballero/portfolio/src/main/webapp/script.js

PR #58, opened 2020-07-13
7/13 Integrate Translate API into Comments Page

annemariecabs:
Use Translate API to translate text of the comments. Add HTML select element to allow user to choose language. Build translate function into the Comment class. Modify getComments() function to allow for necessary query tags. 

Note: This is deployed, so you should be able to check this out.


agcaballero/portfolio/src/main/java/com/google/sps/servlets/DataServlet.java
```diff
@@ -73,15 +76,29 @@ public String toString() {
...
+    }
+
+    /**
+     * Translate the text to the language of a given language code
+     */
+    public String translateText(String languageCode) {
+      // Do the translation.
+      Translate translate = TranslateOptions.getDefaultInstance().getService();
+      Translation translation =
+          translate.translate(text, Translate.TranslateOption.targetLanguage(languageCode));
```
ykjchen, line 98: Is the translate service able to figure out what the current language is? For example, if text is "red", then it's translated to French "rouge", and then translated to some other language, how does the Translate service know the current text is French and not English (it would mean the different between "red" and a type of makeup).

agcaballero/portfolio/src/main/webapp/comments.html
```diff
@@ -51,14 +53,120 @@ <h1>Leave a Comment</h1>
...
+              <option value="TK">Turkmen</option>
+              <option value="UK">Ukrainian</option>
+              <option value="UR">Urdu</option>
+              <option value="UZ">Uzbek</option>
+              <option value="VI">Vietnamese</option>
+              <option value="CY">Welsh</option>
+              <option value="XH">Xhosa</option>
+              <option value="YI">Yiddish</option>
+              <option value="YO">Yoraba</option>
+              <option value="ZU">Zulu</option>
```
ykjchen, line 168: Consider starting off with just a few languages that you're able to test for accuracy.

agcaballero/portfolio/src/main/java/com/google/sps/servlets/DataServlet.java
```diff
@@ -73,15 +76,29 @@ public String toString() {
...
 
       return "<div class='comment-div'>"
           + "<p class='date'>" + formatter.print(localTime) + "</p>"
-          + "<p><b>" + name + " (" + email + "):</b> <br><br>" + text + "</p></div>";
+          + "<p><b>" + name + " (" + email + "):</b></p>"
+          + "<p class='comment-text' lang=" + languageCode + ">" + text + "</p></div>";
+    }
+
+    /**
+     * Translate the text to the language of a given language code
```
ykjchen, line 92: Might also be worth pointing out that this synchronously makes a network call, so could be slow (or should this method be made asynchronous?)

agcaballero/portfolio/src/main/java/com/google/sps/servlets/DataServlet.java
annemariecabs: I have tried this experimentally (e.g. entering text in Spanish and English and translating both to another language like French). It will translate the text properly, so it can figure out the current language. I think it just detects the current language.

agcaballero/portfolio/src/main/webapp/comments.html
annemariecabs: I already tested a couple languages for accuracy, which is why I went ahead to add all the languages supported by the Translate API. @ykjchen are there other tests you suggest I run to ensure that this program works?

agcaballero/portfolio/src/main/java/com/google/sps/servlets/DataServlet.java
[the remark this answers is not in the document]
annemariecabs: I will add to the comment to say (e.g. "EN" for English). @ykjchen do you think this is sufficient clarification?

agcaballero/portfolio/src/main/java/com/google/sps/servlets/DataServlet.java
```diff
@@ -73,15 +76,29 @@ public String toString() {
...
+     */
+    public String translateText(String languageCode) {
+      // Do the translation.
+      Translate translate = TranslateOptions.getDefaultInstance().getService();
+      Translation translation =
+          translate.translate(text, Translate.TranslateOption.targetLanguage(languageCode));
+      text = translation.getTranslatedText();
```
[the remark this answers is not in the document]
annemariecabs: I'm going to detect if the text is in the language it's being translated too, and if not, translate it.

agcaballero/portfolio/src/main/java/com/google/sps/servlets/DataServlet.java
```diff
@@ -73,15 +76,29 @@ public String toString() {
...
           + "<p class='date'>" + formatter.print(localTime) + "</p>"
-          + "<p><b>" + name + " (" + email + "):</b> <br><br>" + text + "</p></div>";
+          + "<p><b>" + name + " (" + email + "):</b></p>"
+          + "<p class='comment-text' lang=" + languageCode + ">" + text + "</p></div>";
+    }
+
+    /**
+     * Translate the text to the language of a given language code
+     */
+    public String translateText(String languageCode) {
```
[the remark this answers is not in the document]
annemariecabs: This actually does translate the text value. However, I had the String be returned as a debugging measure, so one would be able to check the result of the translated text. I'll just change the return value to void.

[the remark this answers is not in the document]
annemariecabs: Done

agcaballero/portfolio/src/main/java/com/google/sps/servlets/DataServlet.java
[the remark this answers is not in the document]
annemariecabs: Done

agcaballero/portfolio/src/main/java/com/google/sps/servlets/DataServlet.java
annemariecabs: @ykjchen I think clarifying in the comment it could be slow is sufficient to me. Do you think it should be made asynchronous? I haven't done that with a Java method before.

agcaballero/portfolio/src/main/java/com/google/sps/servlets/DataServlet.java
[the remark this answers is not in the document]
annemariecabs: Actually, I've decided to have it return text because I'm worried if I change the text every time I might get a strange telephone-like effect. I don't think this is happening, but I think it's safest if this is just a function. I'll change the documentation accordingly.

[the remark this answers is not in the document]
annemariecabs: Done

agcaballero/portfolio/src/main/java/com/google/sps/servlets/DataServlet.java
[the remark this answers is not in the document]
ykjchen: Yeah, that helps to know it's a 2 character language code. Ideally, you'd make note of the ISO spec (if it uses one), so that users know what format to pass in here, but for the purposes of this exercise, if that's not readily available, your example would be sufficient.

ykjchen: In the interest of time, I'd go with the comment as the MVP... but for future reference perhaps a CompletableFuture would be appropriate for tasks like this:
https://www.baeldung.com/java-asynchronous-programming#completableFuture

i.e. the method would return CompletableFuture<String> rather than String.

agcaballero/portfolio/src/main/java/com/google/sps/servlets/DataServlet.java
ykjchen: Maybe this wasn't a good example, and one of these would be a better test:
http://transpremium.com/same-word-different-meaning-englishfrench/

For example, if the word is "chair", which means different things in English and French, and you translate to a different language, which meaning would you end up with?

 it seems like one option for maintaining accuracy would be to also set a sourceLanguage in the TranslationOption:
https://googleapis.dev/java/google-cloud-clients/0.110.0-alpha/com/google/cloud/translate/Translate.TranslateOption.html

agcaballero/portfolio/src/main/java/com/google/sps/servlets/DataServlet.java
```diff
@@ -73,15 +76,39 @@ public String toString() {
...
       return "<div class='comment-div'>"
           + "<p class='date'>" + formatter.print(localTime) + "</p>"
-          + "<p><b>" + name + " (" + email + "):</b> <br><br>" + text + "</p></div>";
+          + "<p><b>" + name + " (<a href='mailto:" + email + "'>" + email + "</a>):</b></p>"
+          + "<p class='comment-text' lang=" + languageCode + ">" + getTranslatedText(languageCode) + "</p></div>";
+    }
+
+    /**
+     * @return the text field translated to the language of a given language code (e.g. "EN" for English)
+     * [Makes a network call to Translation API so method could be slow]
```
ykjchen, line 93: In examples I've seen, generic descriptions goes above the tags in JavaDoc. Do you have examples where descriptions go below the tags?

agcaballero/portfolio/src/main/java/com/google/sps/servlets/DataServlet.java
```diff
@@ -73,15 +76,39 @@ public String toString() {
...
+     */
+    public String getTranslatedText(String languageCode) {
+      // declare an instance of translate
+      Translate translate = TranslateOptions.getDefaultInstance().getService();
+
+      // detect the current language of text
+      String currentLanguage = translate.detect(text).getLanguage();
```
ykjchen, line 100: Seems like this will also cost money, and could be avoided by caching the current language code?

agcaballero/portfolio/src/main/webapp/comments.html
ykjchen: In general, it may be a larger maintenance cost to support more languages (users may write in about translations you're not familiar with), but as long as you've taken this into account and are okay with the tradeoff, I think this is fine. Feel free to resolve this comment after reading.

agcaballero/portfolio/src/main/webapp/script.js
```diff
@@ -177,14 +177,14 @@ console.log(constructSearchLink('Fun Search Term, But Not This Part'));
 
 /** Gets comments from data tag and updates "Comments" page with it */
 async function getComments() {
-  let maxComments = document.getElementById("max-comments").value;
-  if(maxComments == "")
-    maxComments = 50; // set default max comments to 50
-  
+  let maxComments = document.getElementById('max-comments').value;
+  if (maxComments == '') maxComments = 50;  // set default max comments to 50
```
aaronegolden: Is this needed, since the Java code also sets a default of 50 in the getRequestParameter method? There's a downside of having defaults like this set in multiple places, which is that if you ever want to change the defaults you've got to go hunt down the different places where they're set.

agcaballero/portfolio/src/main/java/com/google/sps/servlets/DataServlet.java
annemariecabs: That's a great point -- I will move it up above. I think I just accidentally typed it under. Thank you!

agcaballero/portfolio/src/main/java/com/google/sps/servlets/DataServlet.java
annemariecabs: I agree. I think the best course of action would just be to let the user input the language via dropdown they're typing in, and storing that in the comment as the source language.

agcaballero/portfolio/src/main/java/com/google/sps/servlets/DataServlet.java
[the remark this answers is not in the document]
annemariecabs: I'll try and provide the language spec if I can find it.

agcaballero/portfolio/src/main/webapp/comments.html
annemariecabs: Thanks. That clarification makes sense. I just wanted to try and be accessible, so I think I'll keep the expanded languages. However. I appreciate understanding the trade off better.

agcaballero/portfolio/src/main/java/com/google/sps/servlets/DataServlet.java
annemariecabs: This makes sense. I'm planning to store which language it was submitted in via dropdown. Then, not translate it if this is same as current language.

agcaballero/portfolio/src/main/webapp/script.js
annemariecabs: That makes a lot of sense. I will change that (it should be in comments.js file after I change it).

agcaballero/portfolio/src/main/java/com/google/sps/servlets/DataServlet.java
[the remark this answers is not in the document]
annemariecabs: Done (I found the ISO spec)

agcaballero/portfolio/src/main/java/com/google/sps/servlets/DataServlet.java
annemariecabs: I tried it with chair, and it works now because I began storing the source language.

agcaballero/portfolio/src/main/java/com/google/sps/servlets/DataServlet.java

agcaballero/portfolio/src/main/java/com/google/sps/servlets/DataServlet.java

agcaballero/portfolio/src/main/webapp/script.js
annemariecabs: I decided to put the defaults in the JS file, but remove them from the Java file. This way they're only in one place. Placing them in a Java file seems to cause bugs.

agcaballero/portfolio/src/main/java/com/google/sps/servlets/DataServlet.java
annemariecabs: Thank you for the info! If I have time later this week, I might try this out.

ykjchen:
@annemariecabs Does this have the correct base branch? Should the base branch be ac-maps?

PR #72, opened 2020-07-17
7/16: Add the ability to add all optional attendees if it still works for mandatory attendees (& corresponding tests)

annemariecabs:
Edit algorithm so it builds two collections: one that works for all attendees and one that works for mandatory attendees. If there are any in the first group, that is submitted, else (except in some edge cases), the other is submitted. Add the tests to ensure this function works (e.g. add situations in which there are optional participants in the meeting request).


agcaballero/walkthroughs/week-5-tdd/project/src/main/java/com/google/sps/FindMeetingQuery.java
```diff
@@ -33,54 +33,122 @@
    */ 
   public Collection<TimeRange> query(Collection<Event> events, MeetingRequest request) {
```
ykjchen, line 34: Would it simplify this change if you added a `Collection<Strings> attendees` parameter to this function (i.e. move the existing code in this function to a new function with 3 parameters instead of 2)? Then from this method you could call into that function with both sets of attendees (optional+required and required-only).

annemariecabs: @ykjchen  I see what you're saying. My original reasoning was that this method only requires creating the minutes array once and processing it once. The code is less simple, but more efficient. Do you think it's worth the trade off in efficiency to change it? If so, I'm happy to make the change.

agcaballero/walkthroughs/week-5-tdd/project/src/main/java/com/google/sps/FindMeetingQuery.java
```diff
@@ -33,54 +33,122 @@
...
-
-    for(int i = 0; i < minutes.length; i++) {
-      minutes[i] = true;
+    Collection<String> optionalAttendees = request.getOptionalAttendees();
+    
+    // are there any mandatory employees?
+    boolean mandatory = false;
+    if(attendees.size() > 0) {
+      mandatory = true;
+    }
```
ykjchen, line 42: I think these lines could be condensed to:

```
boolean mandatory = (attendees.size() > 0);
```

Also applies below.

agcaballero/walkthroughs/week-5-tdd/project/src/main/java/com/google/sps/FindMeetingQuery.java
```diff
@@ -33,54 +33,122 @@
...
-    ArrayList<TimeRange> times = new ArrayList<TimeRange>();
-     // minutes of day is + 1 to account for first and last minute of day
-    boolean[] minutes = new boolean[MINUTES_IN_DAY + 1];
-
-    for(int i = 0; i < minutes.length; i++) {
-      minutes[i] = true;
+    Collection<String> optionalAttendees = request.getOptionalAttendees();
+    
+    // are there any mandatory employees?
+    boolean mandatory = false;
```
ykjchen, line 39: Consider naming `hasMandatoryAttendees` for clarity.

agcaballero/walkthroughs/week-5-tdd/project/src/main/java/com/google/sps/FindMeetingQuery.java
```diff
@@ -33,54 +33,122 @@
...
+    if(attendees.size() > 0) {
+      mandatory = true;
+    }
+    
+    // are there any optional employees?
+    boolean optional = false;
```
ykjchen, line 45: Consider naming `hasOptionalAttendees` for clarity.

agcaballero/walkthroughs/week-5-tdd/project/src/main/java/com/google/sps/FindMeetingQuery.java
```diff
@@ -33,54 +33,122 @@
...
+    // 0 - all attendees (mandatory & optional) can attend 
+    // 1 - mandatory attendees can attend (but not all optional)
+    // 2 - not all mandatory attendees can attend (even if all optional can)
+
     for(Event event: events) {
-      // if there's an overlap in attendees block off those times
-      if(attendeeOverlap(event.getAttendees(), attendees)) {
+      int status = 0;
+      
+      if (attendeeOverlap(event.getAttendees(), attendees)) {
```
ykjchen, line 59: Suggested name for more grammatical reading: hasOverlappingAttendees()

agcaballero/walkthroughs/week-5-tdd/project/src/main/java/com/google/sps/FindMeetingQuery.java
```diff
@@ -33,54 +33,122 @@
...
+        status = 2;
+      }
+      else if (attendeeOverlap(event.getAttendees(), optionalAttendees)) {
```
ykjchen, line 62: Style requires K&R style for braces:
https://google.github.io/styleguide/javaguide.html#s4.1.2-blocks-k-r-style

```
} else if (
```

Also applies to other instances in this PR.

agcaballero/walkthroughs/week-5-tdd/project/src/main/java/com/google/sps/FindMeetingQuery.java
annemariecabs: That does sound better. Thank you!

agcaballero/walkthroughs/week-5-tdd/project/src/main/java/com/google/sps/FindMeetingQuery.java
annemariecabs: Thanks for the elaboration. I think I'll try and make the change. There will be a slight decrease in efficiency; however, I think it should be pretty minimal to just add an extra sweep of the array.

agcaballero/walkthroughs/week-5-tdd/project/src/main/java/com/google/sps/FindMeetingQuery.java
```diff
@@ -33,54 +33,122 @@
...
     Collection<String> attendees = request.getAttendees();
-    ArrayList<TimeRange> times = new ArrayList<TimeRange>();
-     // minutes of day is + 1 to account for first and last minute of day
-    boolean[] minutes = new boolean[MINUTES_IN_DAY + 1];
-
-    for(int i = 0; i < minutes.length; i++) {
-      minutes[i] = true;
+    Collection<String> optionalAttendees = request.getOptionalAttendees();
+    
+    // are there any mandatory employees?
```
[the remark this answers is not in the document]
annemariecabs: Yes I did.  I will change that.

agcaballero/walkthroughs/week-5-tdd/project/src/main/java/com/google/sps/FindMeetingQuery.java
```diff
@@ -33,54 +33,122 @@
...
+    // declare variables necessary to keep track of times that work for everyone
+    // and times that work just for mandatory employees
+    int start = 0, startWithOptional = 0; 
+    boolean available = false, availableWithOptional = false;
+    ArrayList<TimeRange> times = new ArrayList<TimeRange>();
+    ArrayList<TimeRange> timesWithOptional = new ArrayList<TimeRange>();
 
-    int start = 0; 
-    boolean available = minutes[start];
+    switch(minutes[start]) {
```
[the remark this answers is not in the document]
annemariecabs: I will fix this (and the other instances of this).

agcaballero/walkthroughs/week-5-tdd/project/src/main/java/com/google/sps/FindMeetingQuery.java
```diff
@@ -33,54 +33,122 @@
...
+      addMeeting(request, start, TimeRange.END_OF_DAY, true, times);
+    }
     
-    return times;
+    // if there are any times where everyone can attend, return those
+    // else just return times where all mandatory attendees can come
+    if(timesWithOptional.size() > 0 || (! mandatory && optional)) {
+      return timesWithOptional;
+    }
+    else  {
```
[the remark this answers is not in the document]
annemariecabs: Thank you for pointing this out! I'll make the change.

agcaballero/walkthroughs/week-5-tdd/project/src/main/java/com/google/sps/FindMeetingQuery.java
annemariecabs: I agree. I was about to make that change myself.

agcaballero/walkthroughs/week-5-tdd/project/src/main/java/com/google/sps/FindMeetingQuery.java
annemariecabs: I changed it. Thank you!

agcaballero/walkthroughs/week-5-tdd/project/src/main/java/com/google/sps/FindMeetingQuery.java
annemariecabs: I will change that. Thank you!

agcaballero/walkthroughs/week-5-tdd/project/src/main/java/com/google/sps/FindMeetingQuery.java
```diff
@@ -33,54 +33,122 @@
...
+    boolean optional = false; 
+    if(optionalAttendees.size() > 0) {
+      optional = true;
     }
 
+    // minutes of day is + 1 to account for first and last minute of day
+    int[] minutes = new int[MINUTES_IN_DAY + 1];
+    // 0 - all attendees (mandatory & optional) can attend 
+    // 1 - mandatory attendees can attend (but not all optional)
+    // 2 - not all mandatory attendees can attend (even if all optional can)
```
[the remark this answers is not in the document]
annemariecabs: That will probably make the code a lot more understandable. I will do that. Thanks!

agcaballero/walkthroughs/week-5-tdd/project/src/main/java/com/google/sps/FindMeetingQuery.java
annemariecabs: I decided to keep the current algorithm, as it would require a total change of the algorithm in this class to make this change, and I think the current one is slightly more efficient. However, I definitely agree it's less understandable, which is why I chose to adopt the enum, and add more comments to clarify my algorithm. Additionally, this algorithm served as a sort of stepping stone to the last algorithm I ended up with for the challenge, so that's also a part of why I wanted to preserve it.

agcaballero/walkthroughs/week-5-tdd/project/src/main/java/com/google/sps/FindMeetingQuery.java
annemariecabs: Actually, I ended up taking this variable out entirely (as I realized a way it wasn't actually necessary). While I do agree with this and the below suggestions regarding these variables, they are currently deleted. Else, I would have adopted these changes.

agcaballero/walkthroughs/week-5-tdd/project/src/main/java/com/google/sps/FindMeetingQuery.java
annemariecabs: I ended up just replacing with the disjoint() function, but this is a better name.

agcaballero/walkthroughs/week-5-tdd/project/src/main/java/com/google/sps/FindMeetingQuery.java
annemariecabs: That makes sense. Thank you!

annemariecabs: Done

agcaballero/walkthroughs/week-5-tdd/project/src/main/java/com/google/sps/FindMeetingQuery.java
annemariecabs: Done

agcaballero/walkthroughs/week-5-tdd/project/src/main/java/com/google/sps/FindMeetingQuery.java
annemariecabs: Done

agcaballero/walkthroughs/week-5-tdd/project/src/main/java/com/google/sps/FindMeetingQuery.java
annemariecabs: Done

agcaballero/walkthroughs/week-5-tdd/project/src/main/java/com/google/sps/FindMeetingQuery.java
annemariecabs: Done

agcaballero/walkthroughs/week-5-tdd/project/src/main/java/com/google/sps/FindMeetingQuery.java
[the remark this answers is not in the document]
annemariecabs: Done

agcaballero/walkthroughs/week-5-tdd/project/src/main/java/com/google/sps/FindMeetingQuery.java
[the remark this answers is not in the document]
annemariecabs: Done

agcaballero/walkthroughs/week-5-tdd/project/src/main/java/com/google/sps/FindMeetingQuery.java
annemariecabs: Done

agcaballero/walkthroughs/week-5-tdd/project/src/main/java/com/google/sps/FindMeetingQuery.java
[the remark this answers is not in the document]
annemariecabs: Done

agcaballero/walkthroughs/week-5-tdd/project/src/main/java/com/google/sps/FindMeetingQuery.java
```diff
@@ -33,68 +37,126 @@
...
-    boolean[] minutes = new boolean[MINUTES_IN_DAY + 1];
+    Collection<String> optionalAttendees = request.getOptionalAttendees();
 
-    for(int i = 0; i < minutes.length; i++) {
-      minutes[i] = true;
-    }
+    // minutes of day is + 1 to account for first and last minute of day
+    Availability[] minutes = new Availability[MINUTES_IN_DAY + 1];
+    // set all spots to default value of all available
+    Arrays.fill(minutes, Availability.ALL_AVAILABLE);
```
ykjchen, line 45: I like how you replaced the loop with this function! Easier to read!

agcaballero/walkthroughs/week-5-tdd/project/src/main/java/com/google/sps/FindMeetingQuery.java
```diff
@@ -33,68 +37,126 @@
...
 
-    for(Event event: events) {
-      // if there's an overlap in attendees block off those times
-      if(attendeeOverlap(event.getAttendees(), attendees)) {
+    for (Event event: events) {
+      Availability status = Availability.ALL_AVAILABLE;
+      
+      if (! Collections.disjoint(event.getAttendees(), attendees)) {
```
ykjchen, line 50: Style omits the space between `! and the next value`:

```
if (!Collections...)
```

Also applies below.

agcaballero/walkthroughs/week-5-tdd/project/src/main/java/com/google/sps/FindMeetingQuery.java
```diff
@@ -33,68 +37,126 @@
...
+    // minutes of day is + 1 to account for first and last minute of day
+    Availability[] minutes = new Availability[MINUTES_IN_DAY + 1];
+    // set all spots to default value of all available
+    Arrays.fill(minutes, Availability.ALL_AVAILABLE);
 
-    for(Event event: events) {
-      // if there's an overlap in attendees block off those times
-      if(attendeeOverlap(event.getAttendees(), attendees)) {
+    for (Event event: events) {
+      Availability status = Availability.ALL_AVAILABLE;
```
ykjchen, line 48: `Availability availability` would make it easier to connect the variable to its type down below (so reader doesn't need to come back here to figure out that `status` is of type `Availability`.

agcaballero/walkthroughs/week-5-tdd/project/src/main/java/com/google/sps/FindMeetingQuery.java
```diff
@@ -33,68 +37,126 @@
...
+
+      if (status != Availability.ALL_AVAILABLE) {
         TimeRange range = event.getWhen();
 
-        for(int i = range.start(); i < range.end(); i++)
-          minutes[i] = false;
+        for (int i = range.start(); i < range.end(); i++)
+          // make sure unavailable times are not overwritten as mandatory available
+          if (minutes[i] != Availability.UNAVAILABLE)
+            minutes[i] = status;
```
ykjchen, line 64: Braces are always required:
https://google.github.io/styleguide/javaguide.html#s4.1.1-braces-always-used

agcaballero/walkthroughs/week-5-tdd/project/src/main/java/com/google/sps/FindMeetingQuery.java
```diff
@@ -33,68 +37,126 @@
...
+      
+      if (! Collections.disjoint(event.getAttendees(), attendees)) {
+        // if there's overlap in mandatory attendees, time is unavailable
+        status = Availability.UNAVAILABLE;
+      } else if (! Collections.disjoint(event.getAttendees(), optionalAttendees)) {
+        // if there's an overlap in only the optional attendees, time is available for only the mandatory employees
+        status = Availability.MANDATORY_AVAILABLE;
+      }
+
+      if (status != Availability.ALL_AVAILABLE) {
```
ykjchen, line 58: This block of code will need to be updated if the enum gets additional values.

Maybe it would be safer (less likely to forget to update this block) to use an exhaustive switch, so that if a new enum value is added, it will also need to be addressed here:

```
switch (status) {
  case  ALL_AVAILABLE:
    continue;
  case UNAVAILABLE:
    Arrays.fill(minutes, range.start(), range.end(), UNAVAILABLE);
    break;
  case MANDATORY_AVAILABLE:
    ...
    break;
}
```

agcaballero/walkthroughs/week-5-tdd/project/src/main/java/com/google/sps/FindMeetingQuery.java
```diff
@@ -33,68 +37,126 @@
...
         TimeRange range = event.getWhen();
 
-        for(int i = range.start(); i < range.end(); i++)
-          minutes[i] = false;
+        for (int i = range.start(); i < range.end(); i++)
+          // make sure unavailable times are not overwritten as mandatory available
+          if (minutes[i] != Availability.UNAVAILABLE)
```
ykjchen, line 63: This check also seems like it would need to be updated if a new enum value is added.

Consider encapsulating this type of check within an enum method:

```
private enum Availability { 
  ALL_AVAILABLE,
  MANDATORY_AVAILABLE,
  UNAVAILABLE
  
  boolean isReplaceableBy(Availability availability) {
    switch (this) {
      case ALL_AVAILABLE:
        return true;
      case UNAVAILABLE:
        return false;
      case MANDATORY_AVAILABLE:
        return availability == ALL_AVAILABLE;
    }
  }
}
```

Then this check becomes:

```
  if (minutes[i].isReplaceableBy(status)) {
    minutes[i] = status;
  }
```

This would make the reason for this check clearer, and also be more resilient to changes to the enum.

agcaballero/walkthroughs/week-5-tdd/project/src/main/java/com/google/sps/FindMeetingQuery.java
```diff
@@ -33,68 +37,126 @@
...
-            
-      }
-      // if current time has been taken until now
-      else {
-        // if now available, set start to now & available to true
-        if(minutes[i]) {
-          start = i;
-          available = true;
-        }
+      switch(minutes[i]) {
```
ykjchen, line 87: Needs space after `switch`

agcaballero/walkthroughs/week-5-tdd/project/src/main/java/com/google/sps/FindMeetingQuery.java
```diff
@@ -33,68 +37,126 @@
...
+        case ALL_AVAILABLE:
+          // start availability period for all attendees (including optional)
+          if (! availableWithOptional) {
```
ykjchen, line 90: Style omits space after `!`. Also applies below.

agcaballero/walkthroughs/week-5-tdd/project/src/main/java/com/google/sps/FindMeetingQuery.java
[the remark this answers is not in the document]
annemariecabs: Done

agcaballero/walkthroughs/week-5-tdd/project/src/main/java/com/google/sps/FindMeetingQuery.java
annemariecabs: Thank you!

agcaballero/walkthroughs/week-5-tdd/project/src/main/java/com/google/sps/FindMeetingQuery.java
annemariecabs: Will fix this. Thanks!

annemariecabs: Done

agcaballero/walkthroughs/week-5-tdd/project/src/main/java/com/google/sps/FindMeetingQuery.java
annemariecabs: That makes sense. I'll change the variable name.

annemariecabs: Done

agcaballero/walkthroughs/week-5-tdd/project/src/main/java/com/google/sps/FindMeetingQuery.java
annemariecabs: Done

agcaballero/walkthroughs/week-5-tdd/project/src/main/java/com/google/sps/FindMeetingQuery.java
annemariecabs: Done

agcaballero/walkthroughs/week-5-tdd/project/src/main/java/com/google/sps/FindMeetingQuery.java
annemariecabs: Done

agcaballero/walkthroughs/week-5-tdd/project/src/main/java/com/google/sps/FindMeetingQuery.java
annemariecabs: That makes sense. Thank you for clarifying this. This seems like a better practice. I will change it.

annemariecabs: Done

agcaballero/walkthroughs/week-5-tdd/project/src/main/java/com/google/sps/FindMeetingQuery.java
annemariecabs: That makes sense. I will add this.

annemariecabs: Done

agcaballero/walkthroughs/week-5-tdd/project/src/main/java/com/google/sps/FindMeetingQuery.java
```diff
@@ -33,68 +54,136 @@
...
+          
+          // end availability period for mandatory attendees & add a possible meeting time range
+          if (available) {
+            addMeeting(request, start, i, false, times);
+            available = false;
+          }
+
+          break;
       }
     }
```
ykjchen, line 160: Two concerns:
1. there's a fair bit of duplicate code here, I'm wondering if there are ways to reduce the duplication (see the DRY principle: https://en.wikipedia.org/wiki/Don%27t_repeat_yourself);
2. this ends up creating both arrays, even if only one of them is needed; perhaps unnecessary processing could be avoided.

One way may be to encapsulate this block of code within a method that accepts a set of Availability values, like:
```
List<TimeRange> timeRangesFromMinuteAvailabilities(Availability[] minuteAvailabilities, Set<Availability> matchingAvailabilities);
```

Then instead of `switch (minutes[i])`, you'd do `matchingAvailabilities.contains(minuteAvailabilites[i])`.

And below, you could then avoid generating the second array if not needed.

```
List<TimeRange> availableTimeRangesIncludingOptionalAttendees =
    timeRangesFromMinuteAvailabilities(minutes, EnumSet.of ( ALL_AVAILABLE ));
if (availableTimeRangesIncludingOptionalAttendees.size() > 0) {
  return availableTimeRangesIncludingOptionalAttendees;
}
return timeRangesFromMinuteAvailabilities(minutes, EnumSet.of ( ALL_AVAILABLE, MANDATORY_AVAILABLE ));
```

agcaballero/walkthroughs/week-5-tdd/project/src/main/java/com/google/sps/FindMeetingQuery.java
```diff
@@ -17,8 +17,29 @@
...
+  
+    boolean isReplaceableBy(Availability availability) {
+      switch (this) {
+        case ALL_AVAILABLE:
+          return true;
+        case UNAVAILABLE:
+          return false;
+        case MANDATORY_AVAILABLE:
+          return availability == ALL_AVAILABLE;
+        default:
```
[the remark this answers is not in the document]
annemariecabs: I added the ```default``` case  because I wanted the method to be accurate even if there were other types in the enums. I can just remove this though if you think it's best.

agcaballero/walkthroughs/week-5-tdd/project/src/main/java/com/google/sps/FindMeetingQuery.java
annemariecabs: I made these changes. I do think it's good to point out that this slightly changes the efficiency of the method. It now makes potentially two sweeps of the minutes array as opposed to one. However, since this doesn't change the time complexity (and this updated change will potentially save time if there is a time when all attendees can make it by only building the one array), I think it's fair to say the trade-off in readability and reduction of redundancy is worth the change.

annemariecabs: Done

agcaballero/walkthroughs/week-5-tdd/project/src/main/java/com/google/sps/FindMeetingQuery.java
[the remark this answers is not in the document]
annemariecabs: Done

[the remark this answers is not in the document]
ykjchen: I'd be concerned that if a new value was added to the enum, it would be easy to overlook updating this, and return an inaccurate value. Perhaps an exception can be thrown in the default case to avoid that?

agcaballero/walkthroughs/week-5-tdd/project/src/main/java/com/google/sps/FindMeetingQuery.java
```diff
@@ -33,68 +57,106 @@
...
+    // get times where all attendees (including optional) can go
+    List<TimeRange> availableTimesWithOptionalAttendees = availableTimeRanges(minutes, EnumSet.of( Availability.ALL_AVAILABLE ), request.getDuration());
+    
+    // if there's at least one time where all optional attendees can go, 
+    // return the times with optional attendees
+    if (availableTimesWithOptionalAttendees.size() > 0) {
+      return availableTimesWithOptionalAttendees;
+    }
+    // else return the times where mandatory attendees can go
+    else {
```
ykjchen, line 106: K&R style:
```
} else {
  // comments
```

agcaballero/walkthroughs/week-5-tdd/project/src/main/java/com/google/sps/FindMeetingQuery.java
```diff
@@ -33,68 +57,106 @@
...
-      // if current time has been taken until now
-      else {
-        // if now available, set start to now & available to true
-        if(minutes[i]) {
-          start = i;
-          available = true;
-        }
+      // If the last minute was unavailable, but this minute is available, then this is the beginning
+      // of a new available time range, so start will be set to this minute and wasLastMinuteAvailable to true.
+      else if (thisMinuteAvailable) {
```
ykjchen, line 154: K&R style:

```
} else if {
  // comments
```

agcaballero/walkthroughs/week-5-tdd/project/src/main/java/com/google/sps/FindMeetingQuery.java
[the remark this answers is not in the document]
annemariecabs: That makes sense. I will change this. Thanks!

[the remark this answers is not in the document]
annemariecabs: Done

agcaballero/walkthroughs/week-5-tdd/project/src/main/java/com/google/sps/FindMeetingQuery.java
annemariecabs: Done

agcaballero/walkthroughs/week-5-tdd/project/src/main/java/com/google/sps/FindMeetingQuery.java
annemariecabs: Done

PR #75, opened 2020-07-17
Complete Query function with optional attendees - Uday

UdayKalra:
Add functionality for optional attendees to query function.


udaykalra/walkthroughs/week-5-tdd/project/src/main/java/com/google/sps/FindMeetingQuery.java
```diff
@@ -87,12 +118,7 @@ private void removeOverlapsWithTimeRange(
     for (TimeRange currentTimeRange : rangesSnapshot) {
 
       // Check for overlap and adjust currentTimeRange
-      if (currentTimeRange.overlaps(possiblyOverlappingTimeRange)) {
-        timeRanges.remove(currentTimeRange);
-        replaceOverlappedTimeRange(possiblyOverlappingTimeRange, currentTimeRange, timeRanges);
-      }
-
-      if (possiblyOverlappingTimeRange.overlaps(currentTimeRange)) {
+      if ((currentTimeRange.overlaps(possiblyOverlappingTimeRange)) || (possiblyOverlappingTimeRange.overlaps(currentTimeRange))) {
```
aaronegolden, line 121: Nit: Remove extra parenthesis around currentTimeRange.overlaps call and possiblyOverlappingTimeRange.overlaps call.

udaykalra/walkthroughs/week-5-tdd/project/src/main/java/com/google/sps/FindMeetingQuery.java
```diff
@@ -31,23 +30,55 @@
...
+   */
   public Collection<TimeRange> query(Collection<Event> events, MeetingRequest request) {
-    /* Create an output collection with a full day to cut(shorten or split)
+
+    // Perform queries with and without optional attendees.
+    Collection<TimeRange> mandatoryCheck = coreQuery(events, request, false);
+    Collection<TimeRange> optionalCheck = coreQuery(events, request, true);
+
+    // Return optional attendee checks if valid.
+    if ((optionalCheck.size() != 0) || (request.getAttendees().size() == 0)) {
```
aaronegolden, line 44: Do we need to check if getOptionalAttendees().size > 0? In other words, does it make sense to return the optionalCheck range if there were no attendees marked as optional?

UdayKalra: If there are no optional attendees then returning the optionalCheck is still the valid return as the algorithm of the optional check is the same as mandatoryCheck but including optional attendees. If there are none, optionalCheck is equivalent to mandatoryCheck.

PR #76, opened 2020-07-20
7/20: Implement optimization challenge (& corresponding tests)

annemariecabs:
Implement the optimization challenge (in which it returns the times in which the most optional attendees can come) by updating the minutes array as an integer array (with a set positive value that means mandatory attendees unavailable and a set negative value which represents how many optional attendees can't make it). Then, the algorithm collects all the times at a certain status or higher, and once it finds at least one time that is of long enough duration it returns the times from that status level. Also, add tests for the specifics of optimization challenge.


agcaballero/walkthroughs/week-5-tdd/project/src/main/java/com/google/sps/FindMeetingQuery.java
```diff
@@ -35,144 +46,130 @@
...
+    // of optional attendees who can't attend. The +1 is to account for last minute in day
+    int[] minuteStatus = new int[MINUTES_IN_DAY + 1];
 
-    // minutes of day is + 1 to account for first and last minute of day
-    int[] minutes = new int[MINUTES_IN_DAY + 1];
-    // 0 - all attendees (mandatory & optional) can attend 
-    // 1 - mandatory attendees can attend (but not all optional)
-    // 2 - not all mandatory attendees can attend (even if all optional can)
+    // keep track of min status reached 
+    int minStatus = 0;
```
ykjchen, line 57: Reading this variable name and comment above, I'm still not sure what this variable is for. Is there a potentially more descriptive name? Longer is fine, clarity is more important than saving horizontal space.

agcaballero/walkthroughs/week-5-tdd/project/src/main/java/com/google/sps/FindMeetingQuery.java
```diff
@@ -35,144 +46,130 @@
...
-    boolean optional = false; 
-    if(optionalAttendees.size() > 0) {
-      optional = true;
-    }
+    // Holds whether the mandatory attendees can attend & the number of optional attendees 
+    // that can attend. If minuteStatus[number] is equal to MANDATORY_UNAVAILABLE, then mandatory
+    // attendees can't attend at that number. If it's equal to 0, then all attendees can attend.
+    // If it's equal to a negative int, then the abs value of that negative number is the number 
+    // of optional attendees who can't attend. The +1 is to account for last minute in day
+    int[] minuteStatus = new int[MINUTES_IN_DAY + 1];
```
ykjchen, line 54: This type of system is hard to understand/maintain. Is it possible to use a class instead?

```
class Availability {
  boolean areMandatoryAttendeesAllAvailable;
  int numberOfOptionalAttendeesAvailable;
}
```

agcaballero/walkthroughs/week-5-tdd/project/src/main/java/com/google/sps/FindMeetingQuery.java
```diff
@@ -35,144 +46,130 @@
...
+              // it will compound)
+              minuteStatus[i] -=  status;
+
+              // if a new minimum status is found, update the min status champion
+              if (minuteStatus[i] < minStatus) {
+                minStatus = minuteStatus[i];
+              }
+            }
+          }
+        }
```
ykjchen, line 97: My guess is that the purpose of this block of code is to populate the `minuteStatus` array, but I'm pretty confused what's going on. I think maybe clearer variable naming would help. In general favor longer names that provide enough description. May take longer to type, but will speed up review times.

agcaballero/walkthroughs/week-5-tdd/project/src/main/java/com/google/sps/FindMeetingQuery.java
```diff
@@ -35,144 +46,130 @@
...
       int status = 0;
-      
-      if (attendeeOverlap(event.getAttendees(), attendees)) {
-        status = 2;
-      }
-      else if (attendeeOverlap(event.getAttendees(), optionalAttendees)) {
-        status = 1;
+      int mandatoryOverlap = amountOfOverlap(event.getAttendees(), attendees);
+
+      if(mandatoryOverlap > 0) {
```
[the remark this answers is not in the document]
annemariecabs: Done

agcaballero/walkthroughs/week-5-tdd/project/src/main/java/com/google/sps/FindMeetingQuery.java
annemariecabs: It keeps track of the lowest status found so that when you look through the array with a status in mind, it'll know when to stop. For example, if the lowest value in the minuteStatus array is -7, then that champion will be -7, and we'll know that if we don't find availabilities where all seven people can't attend (after checking all the previous higher statuses like -6, -5, -4) then we'll have to return no availability. I changed the name to ```lowestStatusFound``` and updated the comment to hopefully make this more clear.

annemariecabs: Done

agcaballero/walkthroughs/week-5-tdd/project/src/main/java/com/google/sps/FindMeetingQuery.java
annemariecabs: Yes. I will make the change.

annemariecabs: Done

agcaballero/walkthroughs/week-5-tdd/project/src/main/java/com/google/sps/FindMeetingQuery.java
annemariecabs: I think the use of the Availability class and implementation of the other edits should help. Please let me know if it needs more clarification.

annemariecabs: Done

agcaballero/walkthroughs/week-5-tdd/project/src/main/java/com/google/sps/FindMeetingQuery.java
```diff
@@ -35,144 +46,130 @@
...
-      for(String attendeeB: groupB)
-        if(attendeeA.equals(attendeeB)) return true;
+  private int amountOfOverlap(Collection<String> left, Collection<String> right) {
+    int overlap = 0;
 
-    return false;
+    for(String leftElement: left)
+      for(String rightElement: right)
+        if(leftElement.equals(rightElement))
+          overlap++;
```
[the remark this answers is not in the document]
annemariecabs: No, it isn't expected. I changed the strategy I used, so now this method should be accurate. The new strategy is just to preprocess left and right so that you have maps with the number of times an element occurs. Then, if an element is a key for both maps, you add the minimum value to overlap.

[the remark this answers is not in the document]
annemariecabs: Done

agcaballero/walkthroughs/week-5-tdd/project/src/main/java/com/google/sps/FindMeetingQuery.java
```diff
@@ -57,106 +96,186 @@ boolean isReplaceableBy(Availability availability) {
...
+    int overlap = 0;
+
+    // pre-process the collections for the number of times an element occurred
+    Map<String, Integer> leftMap = countOccurrences(left);
+    Map<String, Integer> rightMap = countOccurrences(right);
+
+    // increase overlap by the number of times a given element occurred
+    // in both collections
+    for (String key: leftMap.keySet()) {
+      if(rightMap.containsKey(key)) {
```
ykjchen, line 141: Needs space after `if`. Also applies in method below.

agcaballero/walkthroughs/week-5-tdd/project/src/main/java/com/google/sps/FindMeetingQuery.java
```diff
@@ -47,6 +50,42 @@ boolean isReplaceableBy(Availability availability) {
...
+    public Availability() {
+      areMandatoryAttendeesAllAvailable = true;
+      numberOfOptionalAttendeesUnavailable = 0;
+    }
+
+    /**
+     * If any one (or multiple) of the mandatory attendees can't make it, 
+     * sets the mandatory availability of this time to false.
+     */
+    public void mandatoryAttendeeUnavailable() {
```
ykjchen, line 78: Maybe `setMandatoryAttendeesUnavailable()` so that it's clear this method has a side effect?

agcaballero/walkthroughs/week-5-tdd/project/src/main/java/com/google/sps/FindMeetingQuery.java
```diff
@@ -47,6 +50,42 @@ boolean isReplaceableBy(Availability availability) {
    
   /** the number of minutes in a day */
   private static final int MINUTES_IN_DAY = 24 * 60;
+
+  /**
+   * A private class that holds the necessary elements of availability:
+   * can all mandatory attendees attend and how many optional attendees can't come.
+   */
+  private class Availability {
```
ykjchen, line 58: It's a bit confusing to have this class named the same as the enum above (I'm surprised this is allowed). For example, it could be hard to figure out which one List<Availability> contains. Can the two be merged into the enum, or can they be named differently?

agcaballero/walkthroughs/week-5-tdd/project/src/main/java/com/google/sps/FindMeetingQuery.java
```diff
@@ -57,106 +96,186 @@ boolean isReplaceableBy(Availability availability) {
...
 
-          if (duration >= requestDuration) {
-            times.add(TimeRange.fromStartEnd(start, end - 1, true)); // add time range (inclusive of start & end) 
-          }
-          
-          // avoid an inconsistency b/c in case availableMinutes[availableMinutes.length - 1] is true,
-          // you shouldn't set this to false (& that would happen without this if)
-          if (!thisMinuteAvailable) {
-            wasLastMinuteAvailable = false;
+          if (duration >= request.getDuration()) {
```
ykjchen, line 263: Nit: consider caching getDuration() as a local variable outside of the loop to avoid a method call on each iteration.

agcaballero/walkthroughs/week-5-tdd/project/src/main/java/com/google/sps/FindMeetingQuery.java
```diff
@@ -57,106 +96,186 @@ boolean isReplaceableBy(Availability availability) {
...
+        
+          wasLastMinuteAvailable = false;
+        }      
+      } else {
         // If the last minute was unavailable, but this minute is available, then this is the beginning
         // of a new available time range, so start will be set to this minute and wasLastMinuteAvailable to true.
-
-        start = i;
-        wasLastMinuteAvailable = true;
+        if (currentMinuteAvailable) {
```
ykjchen, line 272: Nit: this can be `} else if () {` instead of `} else { if () {`

agcaballero/walkthroughs/week-5-tdd/project/src/main/java/com/google/sps/FindMeetingQuery.java
```diff
@@ -57,106 +96,186 @@ boolean isReplaceableBy(Availability availability) {
...
+    // increase overlap by the number of times a given element occurred
+    // in both collections
+    for (String key: leftMap.keySet()) {
+      if(rightMap.containsKey(key)) {
+        overlap += Math.min(leftMap.get(key), rightMap.get(key));
+      }
+    }
+
+    return overlap;
+  }
```
ykjchen, line 147: I think this is okay, given the requirements, but it's a bit odd... it doesn't seem like we can be sure that two "Alice"s are indeed the same Alice.

In a real use case, we'd probably want to state that these strings are some unique identifier, so that they could be safely deduped by Set.

agcaballero/walkthroughs/week-5-tdd/project/src/main/java/com/google/sps/FindMeetingQuery.java
annemariecabs: Will add.

agcaballero/walkthroughs/week-5-tdd/project/src/main/java/com/google/sps/FindMeetingQuery.java
annemariecabs: That makes sense. Thanks!

agcaballero/walkthroughs/week-5-tdd/project/src/main/java/com/google/sps/FindMeetingQuery.java
annemariecabs: I will rename the enum to AvailabilityOptions. Thank you for the suggestion!

agcaballero/walkthroughs/week-5-tdd/project/src/main/java/com/google/sps/FindMeetingQuery.java
annemariecabs: I will do this. Thanks!

agcaballero/walkthroughs/week-5-tdd/project/src/main/java/com/google/sps/FindMeetingQuery.java
annemariecabs: Thanks! Will change this.

agcaballero/walkthroughs/week-5-tdd/project/src/main/java/com/google/sps/FindMeetingQuery.java
annemariecabs: That makes sense. It is a bit strange. Thanks for giving me the real use case context, it's good to keep in mind.

agcaballero/walkthroughs/week-5-tdd/project/src/main/java/com/google/sps/FindMeetingQuery.java
annemariecabs: Done

agcaballero/walkthroughs/week-5-tdd/project/src/main/java/com/google/sps/FindMeetingQuery.java
annemariecabs: Done

agcaballero/walkthroughs/week-5-tdd/project/src/main/java/com/google/sps/FindMeetingQuery.java
annemariecabs: Done

agcaballero/walkthroughs/week-5-tdd/project/src/main/java/com/google/sps/FindMeetingQuery.java
annemariecabs: Done

agcaballero/walkthroughs/week-5-tdd/project/src/main/java/com/google/sps/FindMeetingQuery.java
annemariecabs: Done

PR #82, opened 2020-07-23
Add workflow to validate and lint backstory files.

ykjchen:


ykjchen:
It seems to be working:
https://github.com/googleinterns/step174-2020/actions/runs/180193885

PR #83, opened 2020-07-23
7/23: Add HTML/CSS for website & fix error in POM

annemariecabs:
Build out home page and Perspective Demo page. Add CSS styling to both. Import proper fonts from Google fonts. Change artifactID in pom.


ykjchen:
Oh, ignore my comment above (https://github.com/googleinterns/step174-2020/pull/83#pullrequestreview-454444294), I was confused about the content of that PR.

backstory/pom.xml
```diff
@@ -0,0 +1,82 @@
+<project xmlns="http://maven.apache.org/POM/4.0.0"
+    xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+    xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+  <modelVersion>4.0.0</modelVersion>
+
+  <groupId>com.google.sps</groupId>
+  <artifactId>backstory</artifactId>
```
ykjchen, line 7: Seems like most of this file should have been unchanged, except fo this line?

backstory/src/main/webapp/index.html
```diff
@@ -0,0 +1,36 @@
...
+  <body class="vertical-flex-container" id="perspective-page">
+    <div class="content vertical-flex-container" id="perspective-content">
+      <h1 class="header blue-header">Backstory Demo</h1>
+
+      <div class="vertical-flex-container" id="demo-divs">
+        <a href="visionDemo.html"
+          ><div class="demo-div" id="vision-demo-div">
+            <h3>Cloud Vision Demo</h3>
+          </div></a
+        >
```
ykjchen, line 22: I think it would be clearer to close the tag on the same line where possible, or is there guidance otherwise?

```
<a href="visionDemo.html">
  <div class="demo-div" id="vision-demo-div">
    <h3>Cloud Vision Demo</h3>
  </div>
</a>
```

backstory/src/main/webapp/perspective-demo.html
```diff
@@ -0,0 +1,26 @@
+<!DOCTYPE html>
+<html lang="en">
+  <head>
+    <meta charset="UTF-8" />
+    <title>Perspective</title>
+    <link rel="stylesheet" href="style.css" />
+    <link rel="stylesheet" href="https://fonts.googleapis.com/css?family=Montserrat" />
+    <link rel="stylesheet" href="https://fonts.googleapis.com/css?family=Roboto" />
+    <link rel="stylesheet" href="https://fonts.googleapis.com/css?family=Lato" />
+    <script src="script.js"></script>
```
ykjchen, line 10: Is there a potentially more specific name for `script.js` that would better indicate why it's used here?

backstory/src/main/webapp/index.html
annemariecabs: No there's not guidance otherwise. Sometimes ```make pretty``` just does that, and since I usually run out before uploading, it's hard to catch. Thanks for pointing it out. I'll fix it.

backstory/pom.xml
annemariecabs: I agree. It seems like GitHub is struggling to merge them properly. The content is the same. I'll just have to resolve it in merge conflicts.

annemariecabs: Done

backstory/src/main/webapp/perspective-demo.html
annemariecabs: Usually, I just leave a call to ```script.js``` in case we have any methods to be used throughout the website. However, currently that script statement does not even need to be there (there's no methods used in perspective-demo from ```script.js```, so I can just remove the script call.

backstory/src/main/webapp/index.html
annemariecabs: Done

backstory/src/main/webapp/perspective-demo.html
annemariecabs: Done

backstory/pom.xml
ykjchen: Looks like this still needs a merge?

backstory/src/main/webapp/index.html
```diff
@@ -0,0 +1,38 @@
...
+          </div>
+        </a>
+
+        <a href="">
+          <div class="demo-div" id="gpt2-demo-div">
+            <h3>GPT-2 Demo</h3>
+          </div>
+        </a>
+
+
```
ykjchen, line 29: Omit this extra newline?

backstory/src/main/webapp/style.css
```diff
@@ -0,0 +1,106 @@
+/* General */
+
+html,
+body {
+  height: 100%;
```
ykjchen, line 5: Was `make pretty` okay with h ordered above f? If so, maybe the linter is not working.

backstory/src/main/webapp/perspective-demo.html
```diff
@@ -0,0 +1,25 @@
...
+  <body class="vertical-flex-container" id="perspective-page">
+    <div class="content vertical-flex-container" id="perspective-content">
+      <h1 class="header blue-header">Check Toxicity</h1>
+
+      <form class="vertical-flex-container" id="perspective-form" onsubmit="printToxicity()">
+        <textarea id="text-for-analysis" name="text"></textarea>
+        <button type="submit">Submit</button>
+      </form>
+
+      <div id="toxicity-percentage"><p>Percentage goes here</p></div>
```
ykjchen, line 22: We talked about this offline... should the Perspective servlet just return the text, if its toxicity is below some threshold?

If not, what object will be responsible for deciding whether the gpt2 text is okay to show?

backstory/src/main/webapp/index.html
annemariecabs: I can remove this. Thanks.

backstory/src/main/webapp/style.css
annemariecabs: I ran the old ```make pretty``` on this. The new make pretty errors out on this (even after the merge) with: ```** [makefile:15: pretty] Error 127``` because it can't find a file (```xargs: node_modules/stylelint/bin/stylelint.js: No such file or directory```). I ran it right after merging with the master, so I can't be sure.

backstory/src/main/webapp/perspective-demo.html
annemariecabs: At this point for the demo (and since we'll be using this page in future as you discussed to check Perspective is working and the attributes of certain text), I thought it was best to just have it return all of the attribute scores. I figured in the flow of the work, a different servlet file would decide (possibly the filtering servlet file) because we also later wanted to add other filters than Perspective. As this page is intended to let us test what scores Perspective produces for different text, I think it would reduce its use if it only showed scores under a certain percentage. I'm definitely open to changing this though.

backstory/pom.xml
annemariecabs: I merged, so if this is still showing as new, I'm not sure if there is a way to resolve it. I'm happy to do whatever you think is best though.

backstory/src/main/webapp/style.css
annemariecabs: While the linter may not work from the command line, the workflow steps you added pre-upload seem to be working, so that's good.

annemariecabs: Update: this does work from the command-line, I just had to install the stylelint.

annemariecabs: Done

backstory/src/main/webapp/perspective-demo.html
annemariecabs: We discussed this more offline. This servlet will probably be kept in the demo stage, and since we've redesigned the way the projects work (not as many servlets involved anymore), this works for the demo, but this servlet will not be used in the MVP.

annemariecabs: Done

backstory/pom.xml
```diff
@@ -35,7 +35,34 @@
...
+      <version>4.12</version>
+      <scope>test</scope>
+    </dependency>
+
+    <dependency>
+      <groupId>org.mockito</groupId>
+      <artifactId>mockito-all</artifactId>
+      <version>1.9.5</version>
+       <scope>test</scope>
+    </dependency>
```
ykjchen, line 51: Are these dependencies related to this PR?

backstory/pom.xml
ykjchen: Thanks the change looks expected now!

backstory/src/main/webapp/perspective-demo.html
```diff
@@ -0,0 +1,25 @@
+<!DOCTYPE html>
+<html lang="en">
+  <head>
+    <meta charset="UTF-8" />
+    <title>Perspective</title>
+    <link rel="stylesheet" href="style.css" />
+    <link rel="stylesheet" href="https://fonts.googleapis.com/css?family=Montserrat" />
+    <link rel="stylesheet" href="https://fonts.googleapis.com/css?family=Roboto" />
+    <link rel="stylesheet" href="https://fonts.googleapis.com/css?family=Lato" />
+    <script src="perspective.js"></script>
```
ykjchen, line 10: Does this file need to get checked in first (I don't see it in this PR, and also don't see it in the repo yet)?

backstory/pom.xml
annemariecabs: I added dependencies I thought would be useful for the project later on, as I was setting up the file structure. As such, I added mockito (I also later updated the version of mockito), junit and the java client I was using for Perspective. They aren't related to this PR though. @ykjchen I can take them out if you want.

backstory/src/main/webapp/perspective-demo.html
annemariecabs: @ykjchen This file is added in next commit. I cherrypicked this commit from other perspective work, which is why it's here. It will be added in the next commit. I'll just remove it for the moment.

backstory/pom.xml
annemariecabs: I took them out.

annemariecabs: Done

backstory/src/main/webapp/perspective-demo.html
annemariecabs: Done

backstory/src/main/webapp/perspective-demo.html
```diff
@@ -0,0 +1,24 @@
...
+    <link rel="stylesheet" href="https://fonts.googleapis.com/css?family=Montserrat" />
+    <link rel="stylesheet" href="https://fonts.googleapis.com/css?family=Roboto" />
+    <link rel="stylesheet" href="https://fonts.googleapis.com/css?family=Lato" />
+  </head>
+
+  <body class="vertical-flex-container" id="perspective-page">
+    <div class="content vertical-flex-container" id="perspective-content">
+      <h1 class="header blue-header">Check Toxicity</h1>
+
+      <form class="vertical-flex-container" id="perspective-form" onsubmit="printToxicity()">
```
ykjchen, line 16: Does printToxicity() exist? What will happen when the user clicks on this?

If the function does not exist, it would be preferable to remove the reference to it, or remove this element altogether, until the function is added.

annemariecabs: It's not used yet, so I removed it.

annemariecabs: Done

PR #105, opened 2020-07-29
7/28: Build front end for MVP 

annemariecabs:
Add home page where the majority of the Backstory flow will occur.  Style home, demos home, and Perspective demo page. 

Note: this is what the front page looks like for reference.

Screenshot 2020-07-28 at 9 47 10 PM


backstory/src/main/webapp/style.css
```diff
@@ -1,3 +1,17 @@
+/* Color Scheme */
+:root {
+  /* Main Color Palette: Bittersweet, Congo Pink, Blue Yonder, Dark Cornflower Blue & Saffron
+     Light/Dark Highlights: Oxford Blue & Alice Blue
+   */
+  --alice-blue: hsla(206, 100%, 96%, 1);
```
UdayKalra, line 6: Will this validate when trying to go in master? May want to change for now.

backstory/src/main/webapp/style.css
```diff
@@ -1,3 +1,17 @@
+/* Color Scheme */
+:root {
```
UdayKalra, line 2: While I wish we could use these, we should aim to use valid conventions that will satisfy master merges for the MVP.

annemariecabs: As we discussed offline, due to new CSS validator, we should be able to use CSS constants.

.github/workflows/backstory-ci.yml
```diff
@@ -24,7 +24,7 @@ jobs:
       run: find backstory/src/main -iname *.html | xargs node_modules/html-validate/bin/html-validate.js
     - name: Validate CSS
       if: always()
-      run: find backstory/src/main -iname *.css | xargs node_modules/css-validator/bin/css-validator
+      run: find backstory/src/main -iname *.css | xargs node_modules/csstree-validator/bin/validate
```
ykjchen, line 27: For future reference, it would have been great if the validator changes could have been made in a separate PR. It better isolates the change (making it easier to find, revert, etc).

annemariecabs: I realized that after I made the commits. I agree it should have been its own PR. In the future, I'll make that a priority. Thank you!

annemariecabs: Done

PR #119, opened 2020-08-02
Jelares Vision Manager

jelares:


jelares:
I changed the file structure to recreate what we discussed in the pod sync. I added the VisionManager interface, along with an implementation and tests for the implementation.

backstory/src/demo/webapp/vision-demo.js
```diff
@@ -0,0 +1,75 @@
...
+          const imageUrl = analyzedImagesObject[i].imageUrl;
+          const labelsJsonArray = analyzedImagesObject[i].labelsJsonArray;
+
+          imageListElement.appendChild(
+              createListElement(imageUrl, labelsJsonArray));
+        }
+      });
+}
+
+/** @return {Element} containing the analyzed image with its labels */
```
annemariecabs, line 51: While this does return an object of type Element, isn't the data type actually object? I was under the impression the data type went inside the {} e.g. ```@return {object} - an Element containing the analyzed image with its labels```.

backstory/src/main/java/com/google/sps/vision/VisionManager.java
```diff
@@ -0,0 +1,46 @@
...
+  public String getLabelsAsJson();
+
+  /**
+   * Returns the image label description fields for all labels.
+   *
+   * @return list of "description" field for all labels returned by Vision API's
+   *     getLabelAnnotationsList method.
+   */
+  public List<String> getLabelDescriptions();
+}
```
annemariecabs, line 46: Add newline.

backstory/src/main/java/com/google/sps/vision/VisionManager.java
```diff
@@ -0,0 +1,46 @@
...
+ * each represent individual labels.
+ */
+
+public interface VisionManager {
+  /**
+   * Returns the full image label annotations as a Json-formatted string of text.
+   *
+   * @return all image labels from Vision API's getLabelAnnotationsList method, in Json.
+   */
+  public String getLabelsAsJson();
```
annemariecabs, line 37: Do interfaces need a public/private declaration? In my experience, you can leave the access type off, and it'll just be assumed to be public (as it's an interface). We should probably standardize this across the project either way.

backstory/src/test/java/com/google/sps/vision/VisionManagerImplTest.java
```diff
@@ -0,0 +1,218 @@
...
+   *
+   * public List<String> getLabelDescriptions();
+   *  Partition on this: labelAnnotations is empty, or non-empty list of EntityAnnotations.
+   */
+
+  private static final byte[] nullRawImageData = null;
+  private static final byte[] emptyRawImageData = new byte[0];
+  private static final byte[] rawImageData = getBytesFromImageReference(
+      "src/test/java/com/google/sps/vision/data/dogRunningOnBeach.jpg", "jpg");
+  private static final List<EntityAnnotation> emptyLabelAnnotations = new ArrayList<>();
```
annemariecabs, line 60: You could place these instantiations in an ```@Before``` section rather than instantiating them up here.

backstory/src/test/java/com/google/sps/vision/VisionManagerImplTest.java
```diff
@@ -0,0 +1,218 @@
...
+  */
+  @Test(expected = IllegalArgumentException.class)
+  public void constructorEmptyImageData() throws IOException {
+    VisionManager visionManagerNullImage = new VisionManagerImpl(emptyRawImageData);
+  }
+
+  /**
+   * Tests paritions on VisionManagerImpl(byte[] rawImageData);
+   *  rawImageData is non-empty image data.
+  */
```
annemariecabs, line 139: In the example we saw of these comments, the test methods had single-line comments not block comments within the method rather than above. This style works as well though. Do you want to make this the standard across projects and add it to the style guide?

backstory/src/test/java/com/google/sps/vision/VisionManagerImplTest.java
```diff
@@ -0,0 +1,218 @@
...
+  @Test
+  public void constructorRawImageData() {
+    List<EntityAnnotation> labelAnnotations = new ArrayList<>();;
```
annemariecabs, line 142: You declare this list more than once in methods. You could probably set this up in the ```@Before``` tag.

backstory/src/demo/webapp/vision-demo.js
jelares: I can change this in a new PR, but it's out of the scope for this one. This PR only rearranges the file structure to put the demos into their own folder. Sorry about the confusion.

backstory/src/main/java/com/google/sps/vision/VisionManager.java
jelares: I think we should include the public tag. It makes the code easier to understand, as someone unfamiliar with interfaces might  be unaware that the assumption is for methods to be public.

backstory/src/main/java/com/google/sps/vision/VisionManager.java
jelares: Added.

backstory/src/test/java/com/google/sps/vision/VisionManagerImplTest.java
jelares: Thanks for the tip, that's what I ended up doing.

backstory/src/test/java/com/google/sps/vision/VisionManagerImplTest.java
jelares: Yeah, we should do Javadoc comments. That way we could make Javadocs for our test suites, if we wanted to.

jelares: Let's add it to the style guide.

backstory/src/test/java/com/google/sps/vision/VisionManagerImplTest.java
jelares: I did that - thanks for the tip.

backstory/src/demo/webapp/vision-demo.js
annemariecabs: That makes sense.

backstory/src/main/java/com/google/sps/vision/VisionManager.java
annemariecabs: Ok, in that case, I will also include the public tag in my interfaces.

backstory/src/test/java/com/google/sps/vision/VisionManagerImplTest.java
annemariecabs: I added it to the style guide. Quick follow-up: partitions*

backstory/src/test/java/com/google/sps/vision/VisionManagerImplTest.java
annemariecabs: Great!

backstory/src/test/java/com/google/sps/vision/VisionManagerImplTest.java
annemariecabs: Sounds good.

PR #123, opened 2020-08-04
8/4: Add PerspectiveManager interface & stubbed implementation of PerspectiveManager

annemariecabs:
Add PerspectiveManager interface and the custom exception it throws (NoAppropriateStoryException). Add a stubbed implementation of PerspectiveManager which returns a default PerspectiveDecision instance.

### DIFF
--- a/backstory/src/main/java/com/google/sps/perspective/PerspectiveManager.java
+++ b/backstory/src/main/java/com/google/sps/perspective/PerspectiveManager.java
@@ -12,17 +12,18 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-package com.google.sps.data.perspective;
+package com.google.sps.perspective;
 
 /**
  * Handles analysis of stories for content and provides a recommendation on
- * the "best" story. Each instance is responsible for analyzing a fixed set of
- * stories.
+ * the "best" appropriate story (if multiple).
  */
 interface PerspectiveManager {
   /**
-   * Analyzes the stories managed by this instance.
+   * Analyzes the stories passed in to this instance.
+   * 
+   * @param story the story to get the decision for
    * @return An object describing the recommendation resulting from the analysis.
    */
-  PerspectiveDecision analyze(String story);
+  public PerspectiveDecision getDecision(String story);
 }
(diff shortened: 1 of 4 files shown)


backstory/src/main/java/com/google/sps/perspective/PerspectiveManagerImpl.java
```diff
@@ -0,0 +1,79 @@
...
+      AttributeType.LIKELY_TO_REJECT,
+      AttributeType.OBSCENE, 
+      AttributeType.PROFANITY, 
+      AttributeType.SEVERE_TOXICITY,
+      AttributeType.SEXUALLY_EXPLICIT, 
+      AttributeType.SPAM, 
+      AttributeType.THREAT,
+      AttributeType.TOXICITY, 
+      AttributeType.UNSUBSTANTIAL
+    };
```
ykjchen, line 44: Nit: looks unrelated to this PR.

annemariecabs: I just added this because it was important for later, and for the overall implementation of PerspectiveManager interface. I'm hopefully not going to have to make any more changes to the interface.

backstory/src/main/java/com/google/sps/perspective/PerspectiveManager.java
```diff
@@ -0,0 +1,30 @@
...
+ */
+interface PerspectiveManager {
+  /**
+   * Analyzes the passed-in story and returns PerspectiveDecision
+   * containing decision based on analysis.
+   *
+   * @param story The story to be analyzed
+   * @return An object describing the recommendation resulting from the analysis.
+   */
+  public PerspectiveDecision getDecision(String story);
```
ykjchen, line 29: @annemariecabs thinking about this more, `get` is often used for property access, so might connote fast access. Maybe consider renaming to `generateDecision()` or something else that indicates that some processing is required?

annemariecabs: Done

PR #124, opened 2020-08-04
Jelares vision manager revised

jelares:


backstory/pom.xml
```diff
@@ -91,13 +103,6 @@
       <version>1.9.59</version>
     </dependency>
 
-    <dependency>
-      <groupId>junit</groupId>
-      <artifactId>junit</artifactId>
-      <version>4.12</version>
-      <scope>test</scope>
-    </dependency>
```
jelares, line 99: I removed this because it was present twice in the POM.

backstory/src/main/java/com/google/sps/vision/VisionManager.java
```diff
@@ -0,0 +1,46 @@
...
+ *
+ * VisionManager Manages the gathering and packaging of Vision API image analytics.
+ * A VisionManager object represents an image annotated with analytics from Vision API.
+ *
+ * The image is represented as the byte data in the rawImageData byte array, and the annotated
+ * analytics are represented as labelAnnotations, a list of Vision EntityAnnotation objects, which
+ * each represent individual labels.
+ */
+
+public interface VisionManager {
```
UdayKalra, line 31: Nit: Interfaces are by default public so it isn't necessary to tag it so.

backstory/src/main/java/com/google/sps/vision/VisionManager.java
```diff
@@ -0,0 +1,46 @@
...
+  /**
+   * Returns the full image label annotations as a Json-formatted string of text.
+   *
+   * @return all image labels from Vision API's getLabelAnnotationsList method, in Json.
+   */
+  public String getLabelsAsJson();
+
+  /**
```
UdayKalra, line 39: Very clear JavaDoc!

backstory/src/main/java/com/google/sps/vision/VisionManagerImpl.java
```diff
@@ -0,0 +1,176 @@
...
+   * must be non-empty. Once set, labelAnnotations will never change, it is immutable. It must be
+   * non-null.
+   *
+   * Safety from Representation Exposure:
+   * All representation fields are immutable.
+   */
+
+  /** Asserts the rep invariants */
+  private void checkRep() {
+    assert rawImageData != null;
```
UdayKalra, line 59: Instead of back-to-back plain asserts, consider throwing descritive exceptions or tagging on strings to return the reason for failure: ``` assert conn != null : "Connection is null" ```

backstory/src/main/java/com/google/sps/vision/VisionManagerImpl.java
```diff
@@ -0,0 +1,176 @@
...
+   * rawImageData will always represent the same image, it is immutable. It must be non-null. It
+   * must be non-empty. Once set, labelAnnotations will never change, it is immutable. It must be
+   * non-null.
+   *
+   * Safety from Representation Exposure:
+   * All representation fields are immutable.
+   */
+
+  /** Asserts the rep invariants */
+  private void checkRep() {
```
UdayKalra, line 58: Even though this is private, a more descriptive method name should be utilized here, especially because Representation Invariant is not a straight forward keyword - consider expanding on it in the javadoc for the method.

backstory/src/main/java/com/google/sps/vision/VisionManagerImpl.java
```diff
@@ -0,0 +1,176 @@
...
+   * @param rawImageData The raw image data for the image being represented in this VisionManager
+   *     object. Must be non-empty and non-null.
+   */
+  public VisionManagerImpl(byte[] rawImageData) throws IllegalArgumentException, IOException {
+    this.rawImageData = rawImageData;
+
+    if (rawImageData != null && rawImageData.length != 0) {
+      this.labelAnnotations = detectLabelsFromImageBytes(rawImageData);
+    } else {
+      throw new IllegalArgumentException();
```
UdayKalra, line 76: Nit: Potentially add an exception string for description.

backstory/src/main/java/com/google/sps/vision/VisionManagerImpl.java
```diff
@@ -0,0 +1,176 @@
...
+   * preset labels.
+   *
+   * @param rawImageData The raw image data for the image being represented in this VisionManager
+   *     object. Must be non-empty and non-null.
+   * @param labelAnnotations the preset labels to annotate the image with. Must be non-null.
+   */
+  public VisionManagerImpl(byte[] rawImageData, List<EntityAnnotation> labelAnnotations)
+      throws IllegalArgumentException {
+    if (rawImageData == null || labelAnnotations == null || rawImageData.length == 0) {
+      throw new IllegalArgumentException();
```
UdayKalra, line 93: Nit: Potentially add an exception string for description.

backstory/src/main/java/com/google/sps/vision/VisionManagerImpl.java
```diff
@@ -0,0 +1,176 @@
...
+    assert rawImageData.length != 0;
+  }
+
+  /**
+   * Instantiates the VisionManager object parameterized with a byte array of raw image data.
+   *
+   * @param rawImageData The raw image data for the image being represented in this VisionManager
+   *     object. Must be non-empty and non-null.
+   */
+  public VisionManagerImpl(byte[] rawImageData) throws IllegalArgumentException, IOException {
```
UdayKalra, line 70: I don't see IOException being thrown?

backstory/src/main/java/com/google/sps/vision/VisionManagerImpl.java
```diff
@@ -0,0 +1,176 @@
...
+    Image img = Image.newBuilder().setContent(ByteString.copyFrom(bytes)).build();
+    Feature feat = Feature.newBuilder().setType(Feature.Type.LABEL_DETECTION).build();
+    AnnotateImageRequest request =
+        AnnotateImageRequest.newBuilder().addFeatures(feat).setImage(img).build();
+    requests.add(request);
+
+    // Initialize the client that will be used to send requests. This client only needs to be
+    // created once, and can be reused for multiple requests. After completing all of the requests
+    // the client will be automatically closed, because it is called within the try block.
+    try (ImageAnnotatorClient client = ImageAnnotatorClient.create()) {
```
UdayKalra, line 158: Is this supposed to catch the IOException?

backstory/src/test/java/com/google/sps/vision/VisionManagerImplTest.java
```diff
@@ -0,0 +1,159 @@
...
+  private byte[] rawImageData;
+  private List<EntityAnnotation> emptyLabelAnnotations;
+  private List<EntityAnnotation> labelAnnotations;
+
+  @Before
+  public void setUp() {
+    nullRawImageData = null;
+    emptyRawImageData = new byte[0];
+    rawImageData = getBytesFromImageReference(
+      "src/test/java/com/google/sps/vision/data/dogRunningOnBeach.jpg", "jpg");
```
UdayKalra, line 69: dogRunningOnBeach, Nice :)

backstory/src/test/java/com/google/sps/vision/VisionManagerImplTest.java
```diff
@@ -0,0 +1,159 @@
...
+   * @return the byte array representation of the image.
+   */
+  private static byte[] getBytesFromImageReference(String reference, String fileType) {
+    ByteArrayOutputStream bos = new ByteArrayOutputStream();
+
+    try {
+      BufferedImage bImage = ImageIO.read(new File(reference));
+      ImageIO.write(bImage, fileType, bos);
+    } catch (IOException exception){
+      System.err.println("IOException while reading image");
```
UdayKalra, line 96: Can the exception be thrown instead of a printed message?

backstory/src/test/java/com/google/sps/vision/VisionManagerImplTest.java
```diff
@@ -0,0 +1,159 @@
...
+    assertEquals(rawImageData, actual.getRawImageData());
+    assertEquals(labelAnnotations, actual.getLabelAnnotations());
+  }
+
+  /**
+   * Tests paritions on getLabelDescriptions();
+   *  labelAnnotations is empty.
+  */
+  @Test
+  public void getLabelDescriptionsEmptyLabelAnnotations() {
```
UdayKalra, line 138: Great Test Naming

backstory/src/main/java/com/google/sps/vision/VisionManager.java
jelares: I talked to AnneMarie about this too, we are going to include the public tag in the interface to make the code easier to understand. In case somebody does not understand that the default for an interface method is that they are public. We could omit it though, if you think there is a good reason for it.

backstory/src/main/java/com/google/sps/vision/VisionManager.java
jelares: Thanks!

backstory/src/test/java/com/google/sps/vision/VisionManagerImplTest.java
jelares: Haha, I only test on images of the highest quality 👍

backstory/src/test/java/com/google/sps/vision/VisionManagerImplTest.java
jelares: Thank you

backstory/src/test/java/com/google/sps/vision/VisionManagerImplTest.java
jelares: Good idea!

jelares: Done.

backstory/src/main/java/com/google/sps/vision/VisionManagerImpl.java
jelares: No, the try opens the client and automatically closes once out of the try block. The reason I did this is to make certain that the client is closed when finished (like using the with statement in python).

backstory/src/main/java/com/google/sps/vision/VisionManagerImpl.java
jelares: Great point, I added strings to explain the assertions.

backstory/src/main/java/com/google/sps/vision/VisionManagerImpl.java
jelares: Sounds good - I added it to the javadoc in our case.

backstory/src/main/java/com/google/sps/vision/VisionManagerImpl.java
jelares: Added.

backstory/src/main/java/com/google/sps/vision/VisionManagerImpl.java
jelares: Added

backstory/src/main/java/com/google/sps/vision/VisionManager.java
UdayKalra: Oh I revised mine to not include public after Joe commented so and that's pushed to the code base at the moment, and AnneMarie's current PR's omit the public so the change may be the move here.

UdayKalra:
Looks good!

backstory/src/test/java/com/google/sps/vision/VisionManagerImplTest.java
```diff
@@ -0,0 +1,159 @@
...
+    VisionManager visionManagerNullImage = new VisionManagerImpl(emptyRawImageData);
+  }
+
+  /**
+   * Tests paritions on VisionManagerImpl(byte[] rawImageData);
+   *  rawImageData is non-empty image data.
+  */
+  @Test
+  public void constructorRawImageData() {
+    VisionManagerImpl actual = new VisionManagerImpl(rawImageData, labelAnnotations);
```
ykjchen, line 128: Does this make a network call? If not, could you point me to where the network call is stubbed out?

backstory/src/main/java/com/google/sps/vision/VisionManagerImpl.java
```diff
@@ -0,0 +1,181 @@
...
+   * Instantiates the VisionManager object parameterized with a byte array of raw image data.
+   *
+   * @param rawImageData The raw image data for the image being represented in this VisionManager
+   *     object. Must be non-empty and non-null.
+   */
+  public VisionManagerImpl(byte[] rawImageData) throws IllegalArgumentException, IOException {
+    this.rawImageData = rawImageData;
+
+    if (rawImageData != null && rawImageData.length != 0) {
+      this.labelAnnotations = detectLabelsFromImageBytes(rawImageData);
```
ykjchen, line 78: I'd recommend not triggering the call to `detectLabels` from the constructor.

Triggering it independently of construction increases the clarity and flexibility of the class:

1. object construction is expected to be fast, and not involve long-running processes,
2. It would be easier to parallelize calls (currently, with the network call being made serially in the constructor, it could be awkward to try to parallelize multiple vision requests),
3. it's possible that this class will adopt other responsibilities, or allow configuration, that can be or need to be done before the network call,
4. in tests, decoupling processing from construction, you would be able to set up the manager in a `setUp()` method.

PR #125, opened 2020-08-04
8/4: Add PerspectiveValues class

annemariecabs:
Add the PerspectiveValue class, which will be a container for text and the resulting values produced by the Perspective API.


backstory/src/main/java/com/google/sps/perspective/PerspectiveManager.java
```diff
@@ -14,11 +14,32 @@
...
 package com.google.sps.perspective;
 
+import au.com.origma.perspectiveapi.v1alpha1.models.AttributeType;
+
 /**
  * Handles analysis of stories for content and provides a recommendation on
  * the "best" appropriate story (if multiple).
  */
-interface PerspectiveManager {
+public interface PerspectiveManager {
```
UdayKalra, line 23: Nit: Interfaces are default public so no need to declare it.
(Thanks Joe ;) )

backstory/src/main/java/com/google/sps/perspective/PerspectiveValue.java
```diff
@@ -0,0 +1,59 @@
...
+
+package com.google.sps.data.perspective;
+
+import au.com.origma.perspectiveapi.v1alpha1.models.AttributeType;
+import java.util.Map;
+
+/**
+ * A container to hold text and the scores generated by that text.
+ */
+public class PerspectiveValue {
```
UdayKalra, line 23: A more descriptive name for this class is probably the move. PerspectiveValue is a little vague considering this is more specifically a container of many perspective values for pre-tagged text. Perhaps PerspectiveValueContainer or PerspectiveValues would be more applicable.

backstory/src/main/java/com/google/sps/perspective/PerspectiveManager.java
annemariecabs: Will change. Thanks!

backstory/src/main/java/com/google/sps/perspective/PerspectiveValue.java
annemariecabs: PerspectiveValues sounds good to me. I'll make the change.

backstory/src/main/java/com/google/sps/perspective/PerspectiveManager.java
annemariecabs: Done

backstory/src/main/java/com/google/sps/perspective/PerspectiveValue.java
annemariecabs: Done

backstory/src/main/java/com/google/sps/perspective/PerspectiveManager.java
```diff
@@ -14,11 +14,32 @@
...
 
+import au.com.origma.perspectiveapi.v1alpha1.models.AttributeType;
+
 /**
  * Handles analysis of stories for content and provides a recommendation on
  * the "best" appropriate story (if multiple).
- */
-interface PerspectiveManager {
+ */ 
+ interface PerspectiveManager {
```
ykjchen, line 23: Indentation?

backstory/src/main/java/com/google/sps/perspective/PerspectiveManager.java
```diff
@@ -14,11 +14,32 @@
...
+      AttributeType.LIKELY_TO_REJECT,
+      AttributeType.OBSCENE, 
+      AttributeType.PROFANITY, 
+      AttributeType.SEVERE_TOXICITY,
+      AttributeType.SEXUALLY_EXPLICIT, 
+      AttributeType.SPAM, 
+      AttributeType.THREAT,
+      AttributeType.TOXICITY, 
+      AttributeType.UNSUBSTANTIAL
+    };
```
ykjchen, line 41: I'm curious why this needed to be in the interface?

backstory/src/main/java/com/google/sps/perspective/PerspectiveValues.java
```diff
@@ -0,0 +1,59 @@
...
+  private final String text;
+
+  /** a map with type-score key-value pairs from the analysis of the text */
+  private final Map<AttributeType, Float> analyses;
+
+  /**
+   * Constructs a PerspectiveValues object with the text that was analyzed
+   * and the subsequent scores produced from Perspective stored in a map.
+   *
+   * @param text the text to analyze
```
ykjchen, line 34: Do you mean `the text that was analyzed`?

backstory/src/main/java/com/google/sps/perspective/PerspectiveValues.java
```diff
@@ -0,0 +1,59 @@
...
+
+/**
+ * A container to hold text and the scores generated by that text.
+ */
+public class PerspectiveValues {
+  /** the text that generated these scores */
+  private final String text;
+
+  /** a map with type-score key-value pairs from the analysis of the text */
+  private final Map<AttributeType, Float> analyses;
```
ykjchen, line 28: What do you think about naming this `attributeTypeToScoreMap`? `analyses` does sound better, but I'm afraid readers will have a hard time figuring out the values mean.

Alternatively, if a map is not strictly needed, and a collection would suffice, you could consider storing this as an array of PerspectiveAnalysis objects for even greater clarity, where:

```
class PerspectiveAnalysis {
  AttributeType attributeType;
  float score;  // would allow use of primitive float
}
```

ykjchen: I see from a later PR that the map helps access to specific scores, so the alternative proposed above probably doesn't apply.

backstory/src/main/java/com/google/sps/perspective/PerspectiveManager.java
annemariecabs: I needed to place these values somewhere so we could have a centralized list of which attributes we were analyzing for. I thought the interface might be a good place because ```PerspectiveManager.REQUIRED_ATTRIBUTES``` seemed like the most logical place they could be put. However, I'm opening to moving this. @ykjchen Do you have a suggestion? Should I place it in ```PerspectiveManagerImpl``` instead?

backstory/src/main/java/com/google/sps/perspective/PerspectiveValues.java
annemariecabs: That is definitely more grammatically correct. Will fix this.

backstory/src/main/java/com/google/sps/perspective/PerspectiveManager.java
annemariecabs: This space was accidental. I'll fix it.

backstory/src/main/java/com/google/sps/perspective/PerspectiveValues.java
annemariecabs: I think that the reader will still understand the attribute type aspect of it because that is the key value. Perhaps I could document this better e.g. ```/** map containing the analysis from the Perspective API as AttributeType keys mapped to their corresponding scores. */ ```. I'm definitely open to changing it but I feel like ```analyses``` emphasizes that this is the results of the Perspective API analysis more so than the alternate name. @ykjchen What do you think?

ykjchen: When writing code devs may reference the documentation to know more about it, but when reading code, it saves time not to need to look it up. For example, if I see somewhere:

```
boolean validate(PerspectiveValues perspectiveValues) {
  for (Map.Entry<AttributeType, Float> analysis : perspectiveValues.analyses.entrySet())  {
    if (analysis.getValue() < 0) {
      return false;
    }
  }
  return true;
}
```

I would wonder why 0 is invalid, because I don't know what the value represents. I'd have to navigate to the class to read the documentation and time == money.

```
boolean validate(PerspectiveValues perspectiveValues) {
  for (Map.Entry<AttributeType, Float> attributeTypeToScoreEntry : perspectiveValues.attributeTypeToScoreMap.entrySet())  {
    if (attributeTypeToScoreEntry.getValue() < 0) {
      return false;
    }
  }
  return true;
}
```

If I see this, I know immediately the value is a score, and it's intuitive that only >0 scores are valid. No navigation to documentation needed!

I'll leave it up to you, but I'd recommend favoring verbosity to potentially save people time in the long run.

backstory/src/main/java/com/google/sps/perspective/PerspectiveManager.java
ykjchen: Do you know yet what objects will need that list? The goal is in general to reduce visibility as much as possible, so that things are exposed only to the classes that need access.

backstory/src/main/java/com/google/sps/perspective/PerspectiveValues.java
annemariecabs: Done

backstory/src/main/java/com/google/sps/perspective/PerspectiveManager.java
annemariecabs: Done

backstory/src/main/java/com/google/sps/perspective/PerspectiveValues.java
annemariecabs: After a few offline conversations, I'm going to go change the name to ``typesToScores``` and update accordingly.

backstory/src/main/java/com/google/sps/perspective/PerspectiveManager.java
annemariecabs: After an offline conversation, as PerspectiveManagerImpl is the only one that needs the class, I will move it there.

backstory/src/main/java/com/google/sps/perspective/PerspectiveValues.java
annemariecabs: Done

backstory/src/main/java/com/google/sps/perspective/PerspectiveManager.java
annemariecabs: Done

PR #126, opened 2020-08-05
8/4: Rename the getDecision() to generateDecision() & add error throw (fix errors from last PR)

annemariecabs:
Implementing [suggestion](https://github.com/googleinterns/step174-2020/pull/123#discussion_r465398150) from PR that had already been merged to rename getDecision() method. Add error throw that was accidentally left out of last PR.


annemariecabs:
> Appears solid. Approving a PR with a TODO is a little odd, perhaps illustrate this is an early implementation for integration purposes.

@UdayKalra I'll add an elaboration that it's for integration purposes.

PR #131, opened 2020-08-05
Jelares Annotated Image Class and Tests

jelares:


backstory/src/main/java/com/google/sps/images/data/AnnotatedImage.java
```diff
@@ -0,0 +1,119 @@
...
+   *
+   * Representation Invariant:
+   * rawImageData will always represent the same image, it is immutable. It must be non-null. It
+   * must be non-empty. Once set, labelAnnotations will never change, it is immutable. It must be
+   * non-null.
+   *
+   * Safety from Representation Exposure:
+   * Representation fields are only returned as immutable representations, or (in the case of
+   * rawImageData) as a defensive copy.
+   */
```
ykjchen, line 50: Make this JavaDoc?

```
/**
 * Class documentation
 */
class ClassName {
...
```

Implementation comments are harder to discover--people reading code don't often navigate to the class unless something's unclear; often IDE's allow you to see documentation just by hovering over a documented thing.

backstory/src/main/java/com/google/sps/images/data/AnnotatedImage.java
```diff
@@ -0,0 +1,119 @@
...
+      throws IllegalArgumentException {
+    if (rawImageData == null || labelAnnotations == null || rawImageData.length == 0) {
+      throw new IllegalArgumentException(
+          "Raw image data must be non-null and non-empty, label annotation data must be non-null");
+    }
+
+    this.rawImageData = rawImageData;
+    this.labelAnnotations = labelAnnotations;
+
+    checkRepresentationInvariantMet();
```
ykjchen, line 81: Is this common practice?

Based on my experiences, I'd favor unit testing this instead, to avoid the need to do this processing in user-facing code.

backstory/src/main/java/com/google/sps/images/data/AnnotatedImage.java
```diff
@@ -0,0 +1,119 @@
...
+  }
+
+  /**
+   * Returns the full image label annotations as a Json-formatted string of text.
+   *
+   * @return all image labels from Vision API's getLabelAnnotationsList method, in Json.
```
ykjchen, line 87: `in Json` -> `as a Json array`?

I'm wondering what the format of the JSON is.

backstory/src/test/java/com/google/sps/images/AnnotatedImageTest.java
```diff
@@ -0,0 +1,159 @@
...
+  @Test
+  public void getLabelDescriptionsEmptyLabelAnnotations() throws IllegalArgumentException {
+    AnnotatedImage annotatedImageActual = new AnnotatedImage(rawImageData, emptyLabelAnnotations);
+    List<String> actual = annotatedImageActual.getLabelDescriptions();
+    List<String> expected = new ArrayList<>();
+    assertEquals(expected, actual);
+  }
+
+  /**
+   * Tests paritions on getLabelDescriptions();
```
ykjchen, line 146: typo: paritions -> partitions (also above)

backstory/src/test/java/com/google/sps/images/AnnotatedImageTest.java
```diff
@@ -0,0 +1,159 @@
...
+
+  /**
+   * public AnnotatedImage(byte[] rawImageData, List<EntityAnnotation> labelAnnotations);
+   *  rawImageData is empty.
+   *  labelAnnotations is non-empty.
+   *  Expects IllegalArgumentException to be thrown.
+  */
+  @Test(expected = IllegalArgumentException.class)
+  public void constructorEmptyImageData() throws IllegalArgumentException {
+    AnnotatedImage annotatedImageActual = new AnnotatedImage(emptyRawImageData, labelAnnotations);
```
ykjchen, line 118: Suggest: `annotatedImageActual` -> `annotatedImage`

1. the last part of the variable name referring to the type is more grammatical: "I have an actual annotated image" vs. "I have an annotated image actual"
2. There is only one instance of AnnotatedImage here, so no disambiguation is needed. Perhaps it would be useful if there was a `actualAnnotatedImage` and `expectedAnnotatedImage`.

Also applies to other test cases.

backstory/src/test/java/com/google/sps/images/AnnotatedImageTest.java
```diff
@@ -0,0 +1,159 @@
...
+  }
+
+  /**
+   * Tests paritions on getLabelDescriptions();
+   *  labelAnnotations is empty.
+  */
+  @Test
+  public void getLabelDescriptionsEmptyLabelAnnotations() throws IllegalArgumentException {
+    AnnotatedImage annotatedImageActual = new AnnotatedImage(rawImageData, emptyLabelAnnotations);
+    List<String> actual = annotatedImageActual.getLabelDescriptions();
```
ykjchen, line 140: Nit: `actual` -> `actualLabelDescriptions` for increased descriptiveness (When I see `actual` I wonder "actual what?")

backstory/src/test/java/com/google/sps/images/AnnotatedImageTest.java
```diff
@@ -0,0 +1,159 @@
...
+    List<String> expected = new ArrayList<>();
```
ykjchen, line 141: Nit: the local variables could be removed for less code and less indirection:

assertEquals(new ArrayList<>(), annotatedImageActual.getLabelDescriptions());

backstory/src/test/java/com/google/sps/images/AnnotatedImageTest.java
```diff
@@ -0,0 +1,159 @@
...
+   *
+   * public List<String> getLabelDescriptions();
+   *  Partition on this: labelAnnotations is empty, or non-empty list of EntityAnnotations.
+   */
+
+  private byte[] nullRawImageData;
+  private byte[] emptyRawImageData;
+  private byte[] rawImageData;
+  private List<EntityAnnotation> emptyLabelAnnotations;
+  private List<EntityAnnotation> labelAnnotations;
```
ykjchen, line 59: Minimizing state in tests is preferred (so reader doesn't need to wonder if this is getting modified by some side effect during a test). If any of these can be converted to constants, or helper methods, those are preferred, e.g.:

```
/** Label annotations with arbitrary elements. */
List<EntityAnnotation> makeNonEmptyLabelAnnotations() {
  List<EntityAnnotation> labelAnnotations = new ArrayList<>();

    EntityAnnotation entityAnnotationMock = mock(EntityAnnotation.class);
    // mock the behavior of stock service to return the value of various stocks
    when(entityAnnotationMock.getDescription()).thenReturn("arbitrary description");  // this description should not be referred to from tests
    labelAnnotations.add(entityAnnotationMock);
    labelAnnotations.add(entityAnnotationMock);
    return labelAnnotations;
}

void testFoo() {
  List<EntityAnnotation> labelAnnotations = makeNonEmptyLabelAnnotations();
  // Do something with labelAnnotations
}
```

backstory/src/test/java/com/google/sps/images/AnnotatedImageTest.java
```diff
@@ -0,0 +1,159 @@
...
+   * Tests paritions on getLabelDescriptions();
+   *  labelAnnotations is a non-empty list of EntityAnnotations.
+  */
+  @Test
+  public void getLabelDescriptionsNonEmptyLabelAnnotations() throws IllegalArgumentException {
+    AnnotatedImage annotatedImageActual = new AnnotatedImage(rawImageData, labelAnnotations);
+    List<String> actual = annotatedImageActual.getLabelDescriptions();
+    List<String> expected = new ArrayList<>();
+    expected.add("DescriptionOne");
+    expected.add("DescriptionOne");
```
ykjchen, line 155: When reading this test, I wonder where "DescriptionOne" comes from. Ideally readers don't have to navigate outside of a test case to understand a test case. Can that be set during this test?

```
String description = "DescriptionOne";
// This would be a private helper
EntityAnnotation entityAnnotation = makeEntityAnnotationWithDescription(description);
List<EntityAnnotation> annotations = ...;
...;
assertEquals(description, annotatedImage.getLabelDescriptions.get(0));
```

This would allow me to read through this without searching for "DescriptionOne".

backstory/src/main/java/com/google/sps/images/data/AnnotatedImage.java
jelares: This is a habit I have from a class about writing safe software. I do not know if it is actually common practice - I could have the method there for testing purposes but not actually call it.

jelares: In theory the rep invariants should never broken.

backstory/src/main/java/com/google/sps/images/data/AnnotatedImage.java
jelares: Sounds good. I will make this javadoc from now on.

backstory/src/main/java/com/google/sps/images/data/AnnotatedImage.java
jelares: It is a JSON array of JSON objects, where each label is represented by a Json object.. I will make this more clear in the documentation.

backstory/src/test/java/com/google/sps/images/AnnotatedImageTest.java
jelares: I updated this piece of documentation to match that of the rest of the test cases.

backstory/src/test/java/com/google/sps/images/AnnotatedImageTest.java
jelares: I renamed it to annotatedImageExpectingException so as to signify that this object creation should raise an exception.

backstory/src/test/java/com/google/sps/images/AnnotatedImageTest.java
jelares: Sounds good! I'll update this and keep it in mind for future testing.

backstory/src/test/java/com/google/sps/images/AnnotatedImageTest.java
jelares: Sure, I did this.

backstory/src/test/java/com/google/sps/images/AnnotatedImageTest.java
jelares: These actually are all constants, I should have declared them final. I will update this.

backstory/src/test/java/com/google/sps/images/AnnotatedImageTest.java
jelares: Sure, I will make a private method to create MockEntityAnnotationWithDescription(description).

backstory/src/test/java/com/google/sps/images/AnnotatedImageTest.java
ykjchen: Oh, actually, if the variable is not used, how about just doing this:

```
public void constructorEmptyImageData() throws IllegalArgumentException {
    new AnnotatedImage(...);
```

That makes it even clearer that the goal of the test is to test some side effect of the constructor (and not its return value).

backstory/src/test/java/com/google/sps/images/AnnotatedImageTest.java
```diff
@@ -0,0 +1,177 @@
...
+  }
+
+  /**
+   * public List<String> getLabelDescriptions();
+   *  labelAnnotations is a non-empty list of EntityAnnotations.
+  */
+  @Test
+  public void getLabelDescriptionsNonEmptyLabelAnnotations() throws IllegalArgumentException, IOException {
+    List<EntityAnnotation> labelAnnotations = new ArrayList<>();
+    labelAnnotations.add(makeMockEntityAnnotationWithDescription("DescriptionOne"));
```
ykjchen, line 159: Consider making this a local variable, to clearly indicate that the assertion has a dependency on this:

```
String firstAnnotationDescription = @"DescriptionOne";
labelAnnotations.add(makeMockEntityAnnotationWithDescription(firstAnnotationDescription));
...
expectedLabelDescriptions.add(firstAnnotationDescription);
```

backstory/src/test/java/com/google/sps/images/AnnotatedImageTest.java
```diff
@@ -0,0 +1,177 @@
...
+
+  /**
+   * public AnnotatedImage(byte[] rawImageData, List<EntityAnnotation> labelAnnotations);
+   *  rawImageData is null.
+   *  labelAnnotations is empty.
+   *  Expects IllegalArgumentException to be thrown.
+  */
+  @Test(expected = IllegalArgumentException.class)
+  public void constructorNullImageData() throws IllegalArgumentException {
+    AnnotatedImage annotatedImageExpectingException = new AnnotatedImage(nullRawImageData, emptyLabelAnnotations);
```
ykjchen, line 95: As mentioned in another comment, may be clearer to just not assign to a local variable, if the variable is unused:

```
public void constructor... {
  new AnnotatedImage(...);
}
```

backstory/src/test/java/com/google/sps/images/AnnotatedImageTest.java
```diff
@@ -0,0 +1,177 @@
...
+  public void getLabelDescriptionsEmptyLabelAnnotations() throws IllegalArgumentException, IOException {
+    byte[] rawImageData;
+    try {
+      rawImageData = getBytesFromImageReference(
+        "src/test/java/com/google/sps/images/data/dogRunningOnBeach.jpg", "jpg");
+    } catch (IOException exception) {
+      throw exception;
+    }
+        
+    AnnotatedImage annotatedImageActual = new AnnotatedImage(rawImageData, emptyLabelAnnotations);
```
ykjchen, line 148: Since there's only one image here, consider naming: `annotatedImage`

The additional "actual" adjective could be useful to differentiate this instance from another AnnotatedImage, but since there is no other, it doesn't contribute to clarity.

ykjchen:
Thanks for the focused PR!

backstory/src/test/java/com/google/sps/images/AnnotatedImageTest.java
jelares: updated!

backstory/src/test/java/com/google/sps/images/AnnotatedImageTest.java
jelares: Updated.

backstory/src/test/java/com/google/sps/images/AnnotatedImageTest.java
jelares: Okay, I did this.

backstory/src/main/java/com/google/sps/images/data/AnnotatedImage.java
```diff
@@ -0,0 +1,118 @@
...
+  /**
+   * Asserts that the representation invariants are met
+   * The representation invariants are the conditions which must always
+   * be true about the class representation fields.
+   */
+  private void checkRepresentationInvariantMet() {
+    assert rawImageData != null : "raw image data field is null";
+    assert labelAnnotations != null : "label annotations data field is null";
+    assert rawImageData.length != 0 : "raw image data field is empty";
+  }
```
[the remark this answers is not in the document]
jelares: I removed the checkRep method, there is really nothing to test that is not already covered by existing test cases.

backstory/src/test/java/com/google/sps/images/AnnotatedImageTest.java
jelares: Updated.

PR #132, opened 2020-08-05
8/4: Add PerspectiveServiceClient and its testing

annemariecabs:
Add PerspectiveServiceClient, which actually calls PerspectiveAPI to analyze text. Add PerspectiveServiceClientTest, which tests PerspectiveServiceClient.

### DIFF
--- a/backstory/src/main/java/com/google/sps/perspective/PerspectiveManager.java
+++ b/backstory/src/main/java/com/google/sps/perspective/PerspectiveManager.java
@@ -39,7 +39,7 @@ interface PerspectiveManager {
       AttributeType.TOXICITY, 
       AttributeType.UNSUBSTANTIAL
     };
-
+    
   /**
    * Analyzes the passed-in story and returns PerspectiveDecision
    * containing decision based on analysis.
--- a/backstory/src/main/java/com/google/sps/perspective/PerspectiveServiceClient.java
+++ b/backstory/src/main/java/com/google/sps/perspective/PerspectiveServiceClient.java
@@ -1,0 +1,96 @@
+// Copyright 2020 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     https://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package com.google.sps.perspective;
+
+import au.com.origma.perspectiveapi.v1alpha1.PerspectiveAPI;
+import au.com.origma.perspectiveapi.v1alpha1.models.AnalyzeCommentRequest;
+import au.com.origma.perspectiveapi.v1alpha1.models.AnalyzeCommentRequest.Builder;
+import au.com.origma.perspectiveapi.v1alpha1.models.AnalyzeCommentResponse;
+import au.com.origma.perspectiveapi.v1alpha1.models.AttributeScore;
+import au.com.origma.perspectiveapi.v1alpha1.models.AttributeType;
+import au.com.origma.perspectiveapi.v1alpha1.models.ContentType;
+import au.com.origma.perspectiveapi.v1alpha1.models.Entry;
+import au.com.origma.perspectiveapi.v1alpha1.models.RequestedAttribute;
+import au.com.origma.perspectiveapi.v1alpha1.models.Score;
+import java.util.HashMap;
+import java.util.Map;
+
+/**
+ * A service client to analyze text using the Perspective API 
+ */
+public class PerspectiveServiceClient {
+
+  /**
+   * Return a PerspectiveValues object with analysis of all of the requested types 
+   * for a specified text using a specified Perspective API.
+   *
+   * @param perspective an instance of the PerspectiveAPI to use to analyze the text for the requested attributeTypes
+   * @param attributeTypes the requested attribute types to analyze the text for
+   * @param text the text to be analyzed by Perspective API
+   * @return an PerspectiveValues object containing all of the scores from the PerspectiveAPI
+   * @throws IllegalArgumentException if any argument is null or if text is empty
+   */
+  public static PerspectiveValues analyze(PerspectiveAPI perspective, AttributeType[] attributeTypes, 
+    String text) {
+    if (perspective == null) {
+      throw new IllegalArgumentException("PerspectiveAPI argument cannot be null.");
+    } else if (attributeTypes == null) {
+      throw new IllegalArgumentException("The array of attribute types cannot be null.");
+    } else if (text == null) {
+      throw new IllegalArgumentException("The text to be analyzed cannot be null");
+    } else if (text.equals("")) {
+      // it can't be empty b/c if it is, when attempting to analyze it with the PerspectiveAPI
+      // a NullPointerException will be thrown 
+      throw new IllegalArgumentException("The text to be analyzed cannot be empty");
+    }
+
+    AnalyzeCommentRequest.Builder builder = new AnalyzeCommentRequest.Builder().comment(
+        new Entry.Builder().type(ContentType.PLAIN_TEXT).text(text).build());
+
+    // add all the attribute types we want to this builder
+    for (AttributeType attributeType : attributeTypes) {
+      builder.addRequestedAttribute(attributeType, null);
+    }
+
+    AnalyzeCommentRequest request = builder.build();
+    AnalyzeCommentResponse response = perspective.analyze(request);
+    
+    Map<AttributeType, Float> analyses = new HashMap<AttributeType, Float>();
+
+    // extract the score for each of the requested attribute types
+    // and put it in a map which will be used to create a PerspectiveValues object
+    for (int i = 0; i < attributeTypes.length; i++) {
+      AttributeType attributeType = attributeTypes[i];
+
+      analyses.put(attributeType, extractScore(response, attributeType));
+    }
+
+    // Convert response to PerspectiveValues object
+    PerspectiveValues analysis = new PerspectiveValues(text, analyses);
+
+    return analysis;
+  }
+
+  /**
+   * helper method to extract the score for a given response & type
+   *
+   * @return the score (as a float) extracted from the response for a given type
+   */
+  private static float extractScore(AnalyzeCommentResponse response, AttributeType type) {
+    AttributeScore attributeScore = response.getAttributeScore(type);
+    Score score = attributeScore.getSummaryScore();
+    return score.getValue();
+  }
+}
--- a/backstory/src/main/java/com/google/sps/perspective/PerspectiveServiceClient.java
+++ b/backstory/src/main/java/com/google/sps/perspective/PerspectiveServiceClient.java
@@ -28,22 +28,22 @@ import java.util.HashMap;
 import java.util.Map;
 
 /**
- * A service client to analyze text using the Perspective API 
+ * A service client to analyze text using the Perspective API
  */
 public class PerspectiveServiceClient {
-
   /**
-   * Return a PerspectiveValues object with analysis of all of the requested types 
+   * Return a PerspectiveValues object with analysis of all of the requested types
    * for a specified text using a specified Perspective API.
    *
-   * @param perspective an instance of the PerspectiveAPI to use to analyze the text for the requested attributeTypes
+   * @param perspective an instance of the PerspectiveAPI to use to analyze the text for the
+   *     requested attributeTypes
    * @param attributeTypes the requested attribute types to analyze the text for
    * @param text the text to be analyzed by Perspective API
    * @return an PerspectiveValues object containing all of the scores from the PerspectiveAPI
    * @throws IllegalArgumentException if any argument is null or if text is empty
    */
-  public static PerspectiveValues analyze(PerspectiveAPI perspective, AttributeType[] attributeTypes, 
-    String text) {
+  public static PerspectiveValues analyze(
+      PerspectiveAPI perspective, AttributeType[] attributeTypes, String text) {
     if (perspective == null) {
       throw new IllegalArgumentException("PerspectiveAPI argument cannot be null.");
     } else if (attributeTypes == null) {
@@ -52,7 +52,7 @@ public class PerspectiveServiceClient {
       throw new IllegalArgumentException("The text to be analyzed cannot be null");
     } else if (text.equals("")) {
       // it can't be empty b/c if it is, when attempting to analyze it with the PerspectiveAPI
-      // a NullPointerException will be thrown 
+      // a NullPointerException will be thrown
       throw new IllegalArgumentException("The text to be analyzed cannot be empty");
     }
 
@@ -66,7 +66,7 @@ public class PerspectiveServiceClient {
 
     AnalyzeCommentRequest request = builder.build();
     AnalyzeCommentResponse response = perspective.analyze(request);
-    
+
     Map<AttributeType, Float> analyses = new HashMap<AttributeType, Float>();
 
     // extract the score for each of the requested attribute types
--- a/backstory/src/test/java/com/google/sps/perspective/PerspectiveServiceClientTest.java
+++ b/backstory/src/test/java/com/google/sps/perspective/PerspectiveServiceClientTest.java
@@ -1,0 +1,188 @@
+// Copyright 2020 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     https://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package com.google.sps;
+
+import static org.mockito.Mockito.*;
+
+import au.com.origma.perspectiveapi.v1alpha1.PerspectiveAPI;
+import au.com.origma.perspectiveapi.v1alpha1.models.AnalyzeCommentRequest;
+import au.com.origma.perspectiveapi.v1alpha1.models.AnalyzeCommentResponse;
+import au.com.origma.perspectiveapi.v1alpha1.models.AttributeScore;
+import au.com.origma.perspectiveapi.v1alpha1.models.AttributeType;
+import au.com.origma.perspectiveapi.v1alpha1.models.Score;
+import com.google.sps.perspective.PerspectiveServiceClient;
+import com.google.sps.perspective.PerspectiveValues;
+import java.util.HashMap;
+import java.util.Map;
+import java.util.Random;
+import org.junit.Assert;
+import org.junit.Before;
+import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.junit.runners.JUnit4;
+import org.mockito.ArgumentCaptor;
+import org.mockito.ArgumentMatchers;
+import org.mockito.invocation.InvocationOnMock;
+import org.mockito.stubbing.Answer;
+
+/** Quality tests for PerspectiveServiceClient */
+@RunWith(JUnit4.class)
+public final class PerspectiveServiceClientTest {
+
+  /** a PerspectiveAPI object to be used as the input api throughout the tests */
+  private static PerspectiveAPI api;
+  /** the array of attribute types that we want scores back for */
+  private static final AttributeType[] desiredTypes = { AttributeType.PROFANITY, AttributeType.TOXICITY, AttributeType.UNSUBSTANTIAL };;
+  /** a Map to be used as the desired scores for PerspectiveAPI to return */
+  private static Map<AttributeType, Float> desiredScores;
+
+
+  @Before 
+  public void setUp() {
+    // create empty or default instances to be used in classes
+    desiredScores = new HashMap<AttributeType, Float>();
+    api = createMockAPI(desiredScores); // will set up an API that only returns -1
+  }
+
+  /**
+   * Call analyze() with a null PerspectiveAPI to ensure 
+   * that an IllegalArgumentException will be thrown.
+   */
+  @Test (expected = IllegalArgumentException.class)
+  public void nullAPIInput() {
+    PerspectiveServiceClient.analyze(null, desiredTypes, "foo");
+  }
+
+  /**
+   * Call analyze() with a null Attribute[] to ensure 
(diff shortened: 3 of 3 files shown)


backstory/src/main/java/com/google/sps/perspective/PerspectiveManager.java
```diff
@@ -39,7 +39,7 @@
       AttributeType.TOXICITY, 
       AttributeType.UNSUBSTANTIAL
     };
-
+
```
annemariecabs, line 42: Accidental space added. Should not affect anything about project. Please ignore.

backstory/src/main/java/com/google/sps/perspective/PerspectiveServiceClient.java
```diff
@@ -0,0 +1,96 @@
...
+   *
+   * @param perspective an instance of the PerspectiveAPI to use to analyze the text for the
+   *     requested attributeTypes
+   * @param attributeTypes the requested attribute types to analyze the text for
+   * @param text the text to be analyzed by Perspective API
+   * @return an PerspectiveValues object containing all of the scores from the PerspectiveAPI
+   * @throws IllegalArgumentException if any argument is null or if text is empty
+   */
+  public static PerspectiveValues analyze(
+      PerspectiveAPI perspective, AttributeType[] attributeTypes, String text) {
```
UdayKalra, line 46: Great edge case handling

annemariecabs: Thanks!

backstory/src/test/java/com/google/sps/perspective/PerspectiveServiceClientTest.java
```diff
@@ -0,0 +1,188 @@
...
+    // set up the next few calls so -1 will be returned (-1 signifies no particular value was desired).
+    when(mockResponse.getAttributeScore(any(AttributeType.class))).thenAnswer(new Answer() {
+      public Object answer(InvocationOnMock invocation) {
+        Object[] args = invocation.getArguments();
+        Object mock = invocation.getMock();
+
+        float fakeScore = -1;
+        AttributeType type = (AttributeType) args[0];
+
+        if (desiredScores.containsKey(type)) {
```
ykjchen, line 175: Maybe move the `float fakeScore = -1` declaration just above this line? Usually it's easier to connect pieces of code when the declaration is close to other references. This is also a strategy for reduce the scope of a variable to just where it's needed (for example, the statement `AttributeType type = (AttributeType) args[0];` doesn't need access to `fakeScore`).

backstory/src/test/java/com/google/sps/perspective/PerspectiveServiceClientTest.java
```diff
@@ -0,0 +1,188 @@
...
+          fakeScore = desiredScores.get(type);
+        }
+
+        when(mockAttributeScore.getSummaryScore()).thenReturn(mockScore);
+        when(mockScore.getValue()).thenReturn(fakeScore);
```
ykjchen, line 180: Would it be safer to create a new instance of Score and AttributeScore each time?

My concern with restubbing a single instance is that if any code holds on to an instance, it would potentially see that instance's score value change over time.

backstory/src/main/java/com/google/sps/perspective/PerspectiveManager.java
annemariecabs: It didn't remove this, but I'll go do it manually.

annemariecabs: Done

backstory/src/main/java/com/google/sps/perspective/PerspectiveServiceClient.java
```diff
@@ -0,0 +1,96 @@
...
+import au.com.origma.perspectiveapi.v1alpha1.models.AttributeType;
+import au.com.origma.perspectiveapi.v1alpha1.models.ContentType;
+import au.com.origma.perspectiveapi.v1alpha1.models.Entry;
+import au.com.origma.perspectiveapi.v1alpha1.models.RequestedAttribute;
+import au.com.origma.perspectiveapi.v1alpha1.models.Score;
+import java.util.HashMap;
+import java.util.Map;
+
+/**
+ * A service client to analyze text using the Perspective API
```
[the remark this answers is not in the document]
annemariecabs: I'll update it to ```A client``` and rename it as well. Thanks!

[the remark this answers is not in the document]
annemariecabs: Done

backstory/src/test/java/com/google/sps/perspective/PerspectiveServiceClientTest.java
```diff
@@ -0,0 +1,188 @@
...
+import org.mockito.ArgumentMatchers;
+import org.mockito.invocation.InvocationOnMock;
+import org.mockito.stubbing.Answer;
+
+/** Quality tests for PerspectiveServiceClient */
+@RunWith(JUnit4.class)
+public final class PerspectiveServiceClientTest {
+
+  /** a PerspectiveAPI object to be used as the input api throughout the tests */
+  private static PerspectiveAPI api;
```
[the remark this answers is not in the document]
annemariecabs: Will rename. Thanks!

backstory/src/main/java/com/google/sps/perspective/PerspectiveServiceClient.java
```diff
@@ -0,0 +1,96 @@
...
+   * for a specified text using a specified Perspective API.
+   *
+   * @param perspective an instance of the PerspectiveAPI to use to analyze the text for the
+   *     requested attributeTypes
+   * @param attributeTypes the requested attribute types to analyze the text for
+   * @param text the text to be analyzed by Perspective API
+   * @return an PerspectiveValues object containing all of the scores from the PerspectiveAPI
+   * @throws IllegalArgumentException if any argument is null or if text is empty
+   */
+  public static PerspectiveValues analyze(
```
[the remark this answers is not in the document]
annemariecabs: That sounds like a good idea. Will change that!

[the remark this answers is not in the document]
annemariecabs: Done

backstory/src/test/java/com/google/sps/perspective/PerspectiveServiceClientTest.java
```diff
@@ -0,0 +1,188 @@
...
+   */
+  @Test
+  public void checkRightScores() {
+    // add the attributes we chose to analyze for and scores to API & create API
+
+    // add each of the desiredTypes and a random float between 0 & 1 as their score 
+    for (AttributeType type: desiredTypes) {
+      Random rand = new Random();
+      
+      desiredScores.put(type, rand.nextFloat());
```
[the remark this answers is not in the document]
annemariecabs: That makes sense. I am just initializing it as a local variable now.

backstory/src/test/java/com/google/sps/perspective/PerspectiveServiceClientTest.java
```diff
@@ -0,0 +1,188 @@
...
+    // when getAttributeScore is called we need to be able to grab the type it is looking for
+    // then, check if this AttributeType is in the map and if it is, grab its value, and
+    // set up the next few calls so that this value will be returned. if it isn't in the map,
+    // set up the next few calls so -1 will be returned (-1 signifies no particular value was desired).
+    when(mockResponse.getAttributeScore(any(AttributeType.class))).thenAnswer(new Answer() {
+      public Object answer(InvocationOnMock invocation) {
+        Object[] args = invocation.getArguments();
+        Object mock = invocation.getMock();
+
+        float fakeScore = -1;
```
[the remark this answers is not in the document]
annemariecabs: That makes sense. I'll make the change.

backstory/src/test/java/com/google/sps/perspective/PerspectiveServiceClientTest.java
annemariecabs: Yes, I see what you mean. I'll move it down.

backstory/src/test/java/com/google/sps/perspective/PerspectiveServiceClientTest.java
annemariecabs: That makes sense. I'll just create a new instance every time.

backstory/src/test/java/com/google/sps/perspective/PerspectiveServiceClientTest.java
```diff
@@ -0,0 +1,188 @@
...
+   * @param desiredScores a map of the scores you want the mockAPI to return 
+   *    (if you don't set any, the mock version of this API will always return -1)
+   * @return the mocked version of the PerspectiveAPI
+   */
+  private static PerspectiveAPI createMockAPI(Map<AttributeType, Float> desiredScores) {
+    // instantiate all the necessary mock objects
+    PerspectiveAPI mockAPI = mock(PerspectiveAPI.class);
+    AnalyzeCommentResponse mockResponse = mock(AnalyzeCommentResponse.class);
+    AttributeScore mockAttributeScore = mock(AttributeScore.class);
+    Score mockScore = mock(Score.class);
```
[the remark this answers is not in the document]
annemariecabs: I see what you mean. I think I'll change it so that the simpler ones (specifically AttributeScore and Score) are just constructed not mocked.

backstory/src/test/java/com/google/sps/perspective/PerspectiveServiceClientTest.java
annemariecabs: Done

backstory/src/test/java/com/google/sps/perspective/PerspectiveServiceClientTest.java
annemariecabs: Done

backstory/src/test/java/com/google/sps/perspective/PerspectiveServiceClientTest.java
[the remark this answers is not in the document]
annemariecabs: Done

backstory/src/test/java/com/google/sps/perspective/PerspectiveServiceClientTest.java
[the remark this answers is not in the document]
annemariecabs: Done

backstory/src/test/java/com/google/sps/perspective/PerspectiveServiceClientTest.java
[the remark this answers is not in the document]
annemariecabs: Done

backstory/src/test/java/com/google/sps/perspective/PerspectiveServiceClientTest.java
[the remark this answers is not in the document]
annemariecabs: Done

backstory/src/main/java/com/google/sps/perspective/PerspectiveServiceClient.java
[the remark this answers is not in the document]
annemariecabs: Done

backstory/src/test/java/com/google/sps/perspective/PerspectiveAPIClientTest.java
```diff
@@ -0,0 +1,191 @@
...
+import org.mockito.stubbing.Answer;
+
+/** Quality tests for PerspectiveAPIClient */
+@RunWith(JUnit4.class)
+public final class PerspectiveAPIClientTest {
+
+  /** an instance of the PerspectiveAPIClient to use throughout our tests */
+  private static PerspectiveAPIClient client;
+  /** a mock PerspectiveAPI object to be used as the input api throughout the tests */
+  private static PerspectiveAPI mockAPI;
```
ykjchen, line 47: Can these two be `private` or `private final` instead of `private static`? JUnit creates a separate instance of this class for each test case, and it would be less error prone to create a new instance for each test, rather than share an instance across multiple tests (in case state gets carried from test to test):
https://stackoverflow.com/a/43730364

backstory/src/test/java/com/google/sps/perspective/PerspectiveAPIClientTest.java
```diff
@@ -0,0 +1,191 @@
...
+    // set up the next few calls so that this value will be returned. if it isn't in the map,
+    // set up the next few calls so -1 will be returned (-1 signifies no particular value was desired).
+    when(mockResponse.getAttributeScore(any(AttributeType.class))).thenAnswer(new Answer() {
+      public Object answer(InvocationOnMock invocation) {
+        Object[] args = invocation.getArguments();
+        Object mock = invocation.getMock();
+
+        AttributeType type = (AttributeType) args[0];
+
+        float scoreValue = -1;
```
ykjchen, line 175: How about setting this to a private constant, so that it seems less like a "magic number" (reader is a bit more aware what the significance of this value is)?

```
private static final DEFAULT_ATTRIBUTE_SCORE = -1;
```

backstory/src/test/java/com/google/sps/perspective/PerspectiveAPIClientTest.java
```diff
@@ -0,0 +1,191 @@
...
+        if (desiredScores.containsKey(type)) {
+          scoreValue = desiredScores.get(type);
+        }
+
+        // create a mocked score with the desired scoreValue and null ScoreType
+        Score mockScore = new Score(scoreValue, null);
```
ykjchen, line 181: `mockScore` -> `score` (since it's a real and not mock object)?

backstory/src/test/java/com/google/sps/perspective/PerspectiveAPIClientTest.java
```diff
@@ -0,0 +1,191 @@
...
+        // create a mocked score with mockScore as summary score and null list of spanscores
+        AttributeScore mockAttributeScore = new AttributeScore(mockScore, null);
```
ykjchen, line 183: `mockAttributeScore` -> `attributeScore` (since it's a real and not mock object)?

backstory/src/test/java/com/google/sps/perspective/PerspectiveAPIClientTest.java
```diff
@@ -0,0 +1,191 @@
...
+/** Quality tests for PerspectiveAPIClient */
+@RunWith(JUnit4.class)
+public final class PerspectiveAPIClientTest {
+
+  /** an instance of the PerspectiveAPIClient to use throughout our tests */
+  private static PerspectiveAPIClient client;
+  /** a mock PerspectiveAPI object to be used as the input api throughout the tests */
+  private static PerspectiveAPI mockAPI;
+  /** the array of attribute types that we want scores back for */
+  private static final AttributeType[] desiredTypes = { AttributeType.PROFANITY, AttributeType.TOXICITY, AttributeType.UNSUBSTANTIAL };;
```
ykjchen, line 49: `Constant names use CONSTANT_CASE`:
https://google.github.io/styleguide/javaguide.html#s5.2.4-constant-names

backstory/src/test/java/com/google/sps/perspective/PerspectiveAPIClientTest.java
```diff
@@ -0,0 +1,191 @@
...
+    Assert.assertEquals(desiredTypes.length, values.getAttributeTypesToScores().size());
+  }
+
+  /**
+   * Check that analyze() returns a PerspectiveValues
+   * object with the same text as passed in.
+   */
+  @Test
+  public void checkTextSame() {
+    PerspectiveValues values = client.analyze(desiredTypes, "foo");
```
ykjchen, line 113: Creating a local variable would make this input more obviously connected to the assertion:

```
String text = "foo";
PerspectiveValues values = client.analyze(desiredTypes, text);
Assert.assertEquals(text, values.getText());
```

backstory/src/test/java/com/google/sps/perspective/PerspectiveAPIClientTest.java
```diff
@@ -0,0 +1,191 @@
...
+  /** a mock PerspectiveAPI object to be used as the input api throughout the tests */
+  private static PerspectiveAPI mockAPI;
+  /** the array of attribute types that we want scores back for */
+  private static final AttributeType[] desiredTypes = { AttributeType.PROFANITY, AttributeType.TOXICITY, AttributeType.UNSUBSTANTIAL };;
+
+
+  @Before 
+  public void setUp() {
+    // will set up an API that only returns -1
+    mockAPI = createMockAPI(new HashMap<AttributeType, Float>());
```
ykjchen, line 55: It doesn't look like this needs to be an instance variable, but can be a local variable instead.

There's only one tests that creates a different PerspectiveAPI mock, and that can be a local variable too.

(This would reduce state within the test case, which reduces room of introducing errors).

backstory/src/test/java/com/google/sps/perspective/PerspectiveAPIClientTest.java
```diff
@@ -0,0 +1,191 @@
...
+   * object with the right scores.
+   */
+  @Test
+  public void checkRightScores() {
+    // a Map to be used as the desired scores for the mock API to return 
+    Map desiredScores = new HashMap<AttributeType, Float>();
+
+    // add each of the desiredTypes and a random float between 0 & 1 as their score 
+    for (AttributeType type: desiredTypes) {
+      Random rand = new Random();
```
ykjchen, line 129: `rand` -> `random`

```
do not abbreviate by deleting letters within a word
```
https://google.github.io/styleguide/jsguide.html#naming-rules-common-to-all-identifiers

backstory/src/test/java/com/google/sps/perspective/PerspectiveAPIClientTest.java
```diff
@@ -0,0 +1,191 @@
...
+import org.junit.runner.RunWith;
+import org.junit.runners.JUnit4;
+import org.mockito.ArgumentCaptor;
+import org.mockito.ArgumentMatchers;
+import org.mockito.invocation.InvocationOnMock;
+import org.mockito.stubbing.Answer;
+
+/** Quality tests for PerspectiveAPIClient */
+@RunWith(JUnit4.class)
+public final class PerspectiveAPIClientTest {
```
ykjchen, line 42: Can you leave a TODO to add a test that checks that the correct AnalyzeCommentRequest is passed to PerspectiveAPI, and follow up with a separate PR with that test case?

backstory/src/test/java/com/google/sps/perspective/PerspectiveAPIClientTest.java
annemariecabs: I agree. I will make them just private.

annemariecabs: Done

backstory/src/test/java/com/google/sps/perspective/PerspectiveAPIClientTest.java
annemariecabs: Thanks for the reminder. This was originally not a constant, but now that it is, I will change it.

annemariecabs: Done

backstory/src/test/java/com/google/sps/perspective/PerspectiveAPIClientTest.java
annemariecabs: That makes sense. Will change this. I kept it because it technically is a part of the mock, but it is real because it's constructed with the public constructor for the class.

backstory/src/test/java/com/google/sps/perspective/PerspectiveAPIClientTest.java
annemariecabs: Will change this. Kept for [same reason](https://github.com/googleinterns/step174-2020/pull/132#discussion_r466814517) as mockScore was kept.

annemariecabs: Done

backstory/src/test/java/com/google/sps/perspective/PerspectiveAPIClientTest.java
annemariecabs: Done

backstory/src/test/java/com/google/sps/perspective/PerspectiveAPIClientTest.java
annemariecabs: That's a great idea. Will definitely make code more comprehensible.

backstory/src/test/java/com/google/sps/perspective/PerspectiveAPIClientTest.java
annemariecabs: I agree. I think it might be even clearer just to have a default text constant throughout file. I think I'll add this:
`/** a piece of default text to use in tests */
  private static final String DEFAULT_TEXT = "foo";`

annemariecabs: Done

annemariecabs:
Note: I did end up changing one of the parameters of analyze() to a List rather than an array because I felt it allowed for more flexibility while maintaining the only requirement (a List with AttributeType objects).

backstory/src/test/java/com/google/sps/perspective/PerspectiveAPIClientTest.java
annemariecabs: Thanks for pointing that out!

annemariecabs: Done

backstory/src/test/java/com/google/sps/perspective/PerspectiveAPIClientTest.java
annemariecabs: Done

backstory/src/test/java/com/google/sps/perspective/PerspectiveAPIClientTest.java
annemariecabs: Great idea! Will do.

annemariecabs: Done

backstory/src/test/java/com/google/sps/perspective/PerspectiveAPIClientTest.java
annemariecabs: That makes sense. I'll make the change.

annemariecabs: Also, I think I'm going to just make client local as well as it takes mockAPI as a parameter.

annemariecabs: Done

PR #137, opened 2020-08-05
8/4: Remove multiples of dependencies from pom.xml

annemariecabs:
Remove duplicate of JUnit dependency tag from pom.xml. Remove outdated mockito tag (there was also an updated mockito dependency tag also present).


backstory/pom.xml
```diff
@@ -50,10 +50,10 @@
       <scope>test</scope>
     </dependency>
 
-    <!-- https://mvnrepository.com/artifact/org.mockito/mockito-inline -->
+    <!-- https://mvnrepository.com/artifact/org.mockito/mockito-core -->
     <dependency>
       <groupId>org.mockito</groupId>
-      <artifactId>mockito-inline</artifactId>
+      <artifactId>mockito-core</artifactId>
```
ykjchen, line 56: The test failure is due to core not supporting mocking final classes.

If you want to use core and mock final classes, you can explicitly enable that feature by including `mock-maker-inline` as done here:
https://github.com/ykjchen/step/pull/1/files

annemariecabs: I will add that. Thanks!

PR #138, opened 2020-08-05
8/5: Add functionality to PerspectiveManager and add PerspectiveManager tests

annemariecabs:
Take out stubbed code and replace with functional code in PerspectiveManager class. Add tests for PerspectiveManager. Add methods to other classes as needed to carry out tests.


backstory/src/main/java/com/google/sps/perspective/PerspectiveManager.java
```diff
@@ -48,10 +55,21 @@
...
    */
-  public PerspectiveManager() throws NoAppropriateStoryException {
-    // TODO: instantiate the PerspectiveAPI for this instance
-    perspectiveAPI = null;
+  public PerspectiveManager() throws APINotAvailableException {
+    PerspectiveAPIFactory factory;
+
+    try {
+      factory = new PerspectiveAPIFactoryImpl();
+    } catch (ClassNotFoundException | NoSuchMethodException | IllegalAccessException
```
UdayKalra, line 67: Blanketing all these different exceptions as a single APINotAvailable exception may frustrating for a user. Perhaps return the specific exception name in the exception string?

backstory/src/test/java/com/google/sps/perspective/PerspectiveManagerTest.java
```diff
@@ -0,0 +1,174 @@
...
+   * to check it constructs a non-null object.
+   */
+  @Test 
+  public void checkNoArgsConstructor() {
+    try {
+      PerspectiveManager manager = new PerspectiveManager();
+      Assert.assertNotNull(manager);
+    }
+    catch(APINotAvailableException exception) {
+      Assert.fail(exception.toString()); // should fail if this exception is thrown
```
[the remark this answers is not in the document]
annemariecabs: Thanks!

backstory/src/main/java/com/google/sps/perspective/PerspectiveManager.java
annemariecabs: I'm including the exception error message in the APINotAvailable error message. @ykjchen Do you think this is sufficient?

backstory/src/main/java/com/google/sps/perspective/PerspectiveManager.java
```diff
@@ -73,10 +91,27 @@ public PerspectiveManager(PerspectiveAPI perspectiveAPI) {
    * @return An object describing the recommendation resulting from the analysis.
    * @throws NoAppropriateStoryException if story is not considered appropriate
    */
-  public StoryDecision generateDecision(String story) {
-    // TODO: Replace this fake behavior with a real API call, etc.
+  public StoryDecision generateDecision(String story) throws NoAppropriateStoryException {
+    PerspectiveAPIClient apiClient = new PerspectiveAPIClient(perspectiveAPI);
+    PerspectiveValues storyValues = apiClient.analyze(REQUESTED_ATTRIBUTES, story);
+    boolean decision = ContentDecisions.makeDecision(storyValues);
```
ykjchen, line 97: IMO `boolean isStoryAppropriate` would be clearer than `boolean decision`.

In conversational language, "Decision? True" is a bit confusing. "Is story appropriate? True" is clearer.

backstory/src/main/java/com/google/sps/perspective/PerspectiveManager.java
```diff
@@ -73,10 +91,27 @@ public PerspectiveManager(PerspectiveAPI perspectiveAPI) {
...
+    throw new NoAppropriateStoryException("The story passed in was not appropriate.");
+  }
+
+  /**
+   * Gets an array of the AttributeTypes that will be analyzed (mainly for testing).
+   *
+   * @return An array of the AttributeTypes we're requesting Perspective to analyze
+   */
+  public static AttributeType[] getRequestedAttributes() {
+    return REQUESTED_ATTRIBUTES;
```
ykjchen, line 115: Since this is a constant, is there any advantage of creating a separate getter? Seems like doing PerspectiveManager.REQUESTED_ATTRIBUTES would serve the same purpose but avoid extra computation.

backstory/src/main/java/com/google/sps/perspective/PerspectiveManager.java
```diff
@@ -48,10 +55,21 @@
...
    * Constructs an object which implements the StoryAnalysisManager
    * class using the Google Perspective API.
+   *
+   * @throws APINotAvailableException when it can't create an instance of the PerspectiveAPI
+   *    (this most likely occurs if the "PerspectiveAPIKey.java" file is not present)
    */
-  public PerspectiveManager() throws NoAppropriateStoryException {
-    // TODO: instantiate the PerspectiveAPI for this instance
-    perspectiveAPI = null;
+  public PerspectiveManager() throws APINotAvailableException {
```
ykjchen, line 62: FYI: For future reference, the link between PerspectiveManager and StoryAnalysisManager would be clearer if this class was named `PerspectiveStoryAnalysisManager` ("a story analysis manager implemented using Perspective").

As is, it's not obvious that PerspectiveManager is related to StoryAnalysisManager unless you check the interface.

backstory/src/main/java/com/google/sps/perspective/data/StoryDecision.java
```diff
@@ -41,4 +41,10 @@ public StoryDecision(String story) throws IllegalArgumentException {
   public String getStory() {
     return story;
   }
+
+  @Override
+  public boolean equals(Object other) {
```
ykjchen, line 46: When overriding equals() it's also recommended to override hashCode():
https://www.geeksforgeeks.org/overriding-equals-method-in-java/

backstory/src/test/java/com/google/sps/perspective/PerspectiveManagerTest.java
```diff
@@ -0,0 +1,174 @@
...
+
+        when(mockAttributeScore.getSummaryScore()).thenReturn(mockScore);
+        when(mockScore.getValue()).thenReturn(fakeScore);
+
+        return mockAttributeScore;
+      }
+    });
+
+    return mockAPI;
+  }
```
ykjchen, line 173: Can this be consolidated into some helper class, to avoid duplication with the PerspectiveAPIClientTests?

backstory/src/test/java/com/google/sps/perspective/PerspectiveManagerTest.java
```diff
@@ -0,0 +1,174 @@
...
+import org.junit.runner.RunWith;
+import org.junit.runners.JUnit4;
+import org.mockito.ArgumentCaptor;
+import org.mockito.ArgumentMatchers;
+import org.mockito.invocation.InvocationOnMock;
+import org.mockito.stubbing.Answer;
+
+/** Quality tests for PerspectiveManager */
+@RunWith(JUnit4.class)
+public final class PerspectiveManagerTest {
```
ykjchen, line 43: Can you add a test case that confirms that:
1. the correct story is passed to perspectiveAPI, and
2. the correct attribute types are set in request passed to the perspectiveAPI?

backstory/src/main/java/com/google/sps/perspective/PerspectiveManager.java
```diff
@@ -48,10 +55,21 @@
...
+    PerspectiveAPIFactory factory;
+
+    try {
+      factory = new PerspectiveAPIFactoryImpl();
+    } catch (ClassNotFoundException | NoSuchMethodException | IllegalAccessException
+        | InvocationTargetException exception) {
+      throw new APINotAvailableException("Perspective API is not available: " + exception);
+    }
+
+    perspectiveAPI = factory.newInstance();
```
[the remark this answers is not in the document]
annemariecabs: That makes sense. Thanks!

backstory/src/main/java/com/google/sps/perspective/PerspectiveManager.java
annemariecabs: That makes sense. I'll change this.

annemariecabs: Done

backstory/src/main/java/com/google/sps/perspective/PerspectiveManager.java
annemariecabs: This is for the purposes of testing (I wanted to check that we're getting the right number of attributes). I can also just make this public. I'm happy to do either. @ykjchen Which would you prefer?

backstory/src/main/java/com/google/sps/perspective/PerspectiveManager.java
annemariecabs: I changed the name, so it should hopefully be clearer.

annemariecabs: Done

backstory/src/main/java/com/google/sps/perspective/data/StoryDecision.java
annemariecabs: I added a simple hashCode() method as well.

annemariecabs: Done

backstory/src/main/java/com/google/sps/perspective/PerspectiveManager.java
annemariecabs: Done

backstory/src/test/java/com/google/sps/perspective/PerspectiveManagerTest.java
annemariecabs: I created a `MockPerspectiveAPIFactory` in the text folder that will manufacture instances like this for my classes.

annemariecabs: Done

backstory/src/test/java/com/google/sps/perspective/PerspectiveManagerTest.java
annemariecabs: I just wrote those tests. Thank you for the suggestion!

annemariecabs: Done

backstory/src/main/java/com/google/sps/perspective/PerspectiveManager.java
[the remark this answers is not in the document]
ykjchen: Gotcha, I didn't realize Java had this limitation. Thanks for explaining.

backstory/src/test/java/com/google/sps/perspective/data/MockPerspectiveAPIFactory.java
```diff
@@ -0,0 +1,117 @@
...
+  }
+
+  /** 
+   * Generates a mock version of the PerspectiveAPI which will return the scores
+   * set in the constructor or which returns a default value when there is no desired score
+   *
+   * @return an instance of a mock Perspective API 
+   */
+  public PerspectiveAPI newInstance() {
+    return perspectiveAPI;
```
ykjchen, line 70: The implication of `newInstance()` is that the caller will get a new instance. This could be resolved here by instead doing: `return createMockAPI(desiredScores);`

backstory/src/test/java/com/google/sps/perspective/PerspectiveStoryAnalysisManagerTest.java
```diff
@@ -0,0 +1,189 @@
...
+  @Test
+  public void checkNoArgsConstructor() {
+    try {
+      PerspectiveStoryAnalysisManager manager = new PerspectiveStoryAnalysisManager();
+      Assert.assertNotNull(manager);
+    }
+    catch(APINotAvailableException exception) {
+      Assert.fail(exception.toString()); // should fail if this exception is thrown
+    }
+  }
```
ykjchen, line 78: When running this test on Github, this will throw since the API key is not checked in. I'm glad you thought of this, but seems reasonable to remove this since its behavior will be different depending on whether you're running this test locally or on Github.

backstory/src/test/java/com/google/sps/perspective/data/MockPerspectiveAPIFactory.java
annemariecabs: Ok I will make that change. Thank you so much!

backstory/src/test/java/com/google/sps/perspective/PerspectiveStoryAnalysisManagerTest.java
annemariecabs: That makes sense. Thank you!

annemariecabs: Done

backstory/src/test/java/com/google/sps/perspective/data/MockPerspectiveAPIFactory.java
annemariecabs: Done

PR #139, opened 2020-08-05
Make StoryManager interfaces public for use outside of package.

UdayKalra:
It appears interfaces without an access tag will assume standard default permission, so public will be needed for access outside its respective folder.


ykjchen:
Thanks for the small PR.

PR #142, opened 2020-08-06
Jelares Images Manager

jelares:


backstory/src/main/java/com/google/sps/perspective/PerspectiveManagerImpl.java
```diff
@@ -72,8 +72,9 @@ public PerspectiveManagerImpl(PerspectiveAPI perspectiveAPI) {
    * @throws NoAppropriateStoryException if no story from decision is considered appropriate
    */
   public PerspectiveDecision generateDecision(String story) throws NoAppropriateStoryException {
-    // TODO: Replace this fake behavior with a real API call, etc. Fake behavior here so teammates have stubbed
-    //       version of this class to use in integration. 
+    // TODO: Replace this fake behavior with a real API call, etc. Fake behavior here so teammates
+    // have stubbed
+    //       version of this class to use in integration.
```
jelares, line 77: Make pretty updated this.

backstory/src/test/java/com/google/sps/images/AnnotatedImageTest.java
```diff
@@ -71,7 +71,7 @@ private static EntityAnnotation makeMockEntityAnnotationWithDescription(String d
    * @param fileType the file extension of the image.
    * @return the byte array representation of the image.
    */
-  private static byte[] getBytesFromImageReference(String reference, String fileType) throws IOException {
+  public static byte[] getBytesFromImageReference(String reference, String fileType) throws IOException {
```
jelares, line 74: Changed this for re-usability (to not have to copy and paste this function for new tests).

backstory/src/main/java/com/google/sps/perspective/PerspectiveManagerImpl.java
ykjchen: Maybe move the string starting with "version" into the line above?

backstory/src/main/java/com/google/sps/images/ImagesManager.java
```diff
@@ -0,0 +1,40 @@
...
+  /**
+   * Runs analytics on a list of images represented by their raw byte data and packages the images
+   * along with their analytics into annotated image objects.
+   *
+   * @param imagesAsByteArrays a list of byte arrays such that each byte array within the list
+   *     represents an image uploaded to
+   * backstory.
+   * @return a list of annotated images correspending to the original images in imagesAsByteArrays.
+   */
+  public List<AnnotatedImage> createAnnotatedImagesFromImagesAsByteArrays(
```
ykjchen, line 38: Thanks for the descriptive name.

backstory/src/main/java/com/google/sps/images/VisionManager.java
```diff
@@ -0,0 +1,101 @@
...
+import java.util.ArrayList;
+import java.util.List;
+
+/**
+ * {@inheritDoc}
+ *
+ * VisionManager is an ImagesManager implemented using the Google Vision API.
+ * VisionManager Manages the gathering and packaging of Vision API image analytics.
+ */
+public final class VisionManager implements ImagesManager {
```
ykjchen, line 36: Nit: naming this VisionImagesManager would make more obvious the link between VisionManager and ImagesManager (VisionImagesManager == an ImagesManager implemented using Vision)

backstory/src/main/java/com/google/sps/images/VisionManager.java
```diff
@@ -0,0 +1,101 @@
...
+  private final ImageAnnotatorClient imageAnnotatorClient;
+
+  /**
+   * Creates a vision manager object using the image annotator client from Vision API.
+   */
+  public VisionManager() throws IOException {
+    ImageAnnotatorClient imageAnnotatorClient = ImageAnnotatorClient.create();
+    this.imageAnnotatorClient = imageAnnotatorClient;
```
ykjchen, line 44: Prefer calling the other constructor, to avoid duplicating code.

```
this(imageAnnotatorClient);
```

backstory/src/main/java/com/google/sps/images/VisionManager.java
```diff
@@ -0,0 +1,101 @@
...
+  public VisionManager(ImageAnnotatorClient imageAnnotatorClient) {
+    this.imageAnnotatorClient = imageAnnotatorClient;
+  }
+
+  @Override
+  public List<AnnotatedImage> createAnnotatedImagesFromImagesAsByteArrays(
+      List<byte[]> imagesAsByteArrays) throws IOException {
+    List<AnnotatedImage> annotatedImages = new ArrayList<>();
+
+    for (byte[] rawImageData : imagesAsByteArrays) {
```
ykjchen, line 59: Maybe add a TODO here to parallelize (its rare to make request serially unless they have an order dependency).

```
// TODO: parallelize calls to detectLabelsFromImageBytes() since it requires a network call.
```

backstory/src/main/java/com/google/sps/images/VisionManager.java
```diff
@@ -0,0 +1,101 @@
...
+    List<AnnotateImageRequest> requests = new ArrayList<>();
+    List<EntityAnnotation> labels;
+
+    Image img = Image.newBuilder().setContent(ByteString.copyFrom(bytes)).build();
+    Feature feat = Feature.newBuilder().setType(Feature.Type.LABEL_DETECTION).build();
+    AnnotateImageRequest request =
+        AnnotateImageRequest.newBuilder().addFeatures(feat).setImage(img).build();
+    requests.add(request);
+
+    // ImageAnnotatorClient will make the requests
```
ykjchen, line 85: Maybe more descriptive:

```
The invocation of batchAnnotateImages() makes a network call.
```

backstory/src/main/java/com/google/sps/images/VisionManager.java
```diff
@@ -0,0 +1,101 @@
...
+    BatchAnnotateImagesResponse response = imageAnnotatorClient.batchAnnotateImages(requests);
+    List<AnnotateImageResponse> responses = response.getResponsesList();
+
+    // For the MVP there is only one image in the batch response.
```
ykjchen, line 89: How about checking and throwing if not... that will avoid this expectation changing out from under you.

Also, this seems less about the MVP and more about the argument accepted by this method.

```
// Annotations for a only a single image are requested, so only a single response is expected.
```

backstory/src/main/java/com/google/sps/images/VisionManager.java
```diff
@@ -0,0 +1,101 @@
...
+    AnnotateImageResponse res = responses.get(0);
+
+    if (res.hasError()) {
+      System.out.format("Error: %s%n", res.getError().getMessage());
```
ykjchen, line 93: There are two format specifiers here, but only one argument... was is the %n for?

backstory/src/main/java/com/google/sps/images/VisionManager.java
```diff
@@ -0,0 +1,101 @@
...
+      return new ArrayList<EntityAnnotation>();
```
ykjchen, line 94: Should this throw instead of swallowing the error? How will the user of this method know whether they should retry detection?

backstory/src/test/java/com/google/sps/images/VisionManagerTest.java
```diff
@@ -0,0 +1,73 @@
...
+/**
+ * Tests for VisionManager.
+ */
+public final class VisionManagerTest {
+
+  /**
+   * Checkts that the request given to Vision API is valid
+   */
+  @Test
+  public void testCorrectRequestInput() throws IOException {
```
ykjchen, line 42: Will you also be testing the output?

For example, in this test you could assign the return value to a List<AnnotatedImages> variable, and check that its contents are expected.

backstory/src/test/java/com/google/sps/images/VisionManagerTest.java
```diff
@@ -0,0 +1,73 @@
...
+    byte[] rawImageData = AnnotatedImageTest.getBytesFromImageReference(
+        "src/test/java/com/google/sps/images/data/dogRunningOnBeach.jpg", "jpg");
+    rawImageDataList.add(rawImageData);
+
+    // Create the necessary mocks
+    ImageAnnotatorClient mockImageAnnotatorClient = mock(ImageAnnotatorClient.class);
+    BatchAnnotateImagesResponse mockBatchResponse = mock(BatchAnnotateImagesResponse.class);
+    List<AnnotateImageResponse> mockResponses = mock(List.class);
+    AnnotateImageResponse mockResponse = mock(AnnotateImageResponse.class);
+    List<EntityAnnotation> mockLabelAnnotations = mock(List.class);
```
ykjchen, line 53: Can real objects be used for any of these? Mocks are typically used as a last resort, when its infeasible to construct real versions of the object.

A real list could be used, and probably real AnnotationImageResponse, etc.

ykjchen:
Thanks for limiting the scope of this PR!

backstory/src/main/java/com/google/sps/images/VisionManager.java
```diff
@@ -0,0 +1,101 @@
...
+    AnnotateImageRequest request =
+        AnnotateImageRequest.newBuilder().addFeatures(feat).setImage(img).build();
+    requests.add(request);
+
+    // ImageAnnotatorClient will make the requests
+    BatchAnnotateImagesResponse response = imageAnnotatorClient.batchAnnotateImages(requests);
+    List<AnnotateImageResponse> responses = response.getResponsesList();
+
+    // For the MVP there is only one image in the batch response.
+    AnnotateImageResponse res = responses.get(0);
```
aaronegolden, line 90: Is it possible for the response list to be empty (e.g. due to an error)? If so, does get(0) throw an exception? Might be worth exercising this case in a unit test just to make sure nothing especially bad happens.

backstory/src/main/java/com/google/sps/images/VisionManager.java
jelares: Good idea, updated.

backstory/src/main/java/com/google/sps/images/ImagesManager.java
jelares: No problem!

backstory/src/main/java/com/google/sps/perspective/PerspectiveManagerImpl.java
jelares: Sure, updated.

backstory/src/main/java/com/google/sps/images/VisionManager.java
jelares: I tried doing this at first and an error was raised - it said that a call to a constructor must be the first line in a method.

backstory/src/main/java/com/google/sps/images/VisionManager.java
jelares: Absolutely right. I was thinking about this, but for the MVP (since we only allow one image upload anyways) we can keep the current implementation. Updated to add the to-do.

backstory/src/main/java/com/google/sps/images/VisionManager.java
jelares: Updated.

backstory/src/main/java/com/google/sps/images/VisionManager.java
jelares: True, good idea - although I want this method to be updated in the future to allow for batch processing. I will do what you mention, then also add the same TO-DO from before.

backstory/src/main/java/com/google/sps/images/VisionManager.java
jelares: Good point, that was a typo.. I am going to update this to match your next comment anyways.

backstory/src/main/java/com/google/sps/images/VisionManager.java
jelares: Updated.

backstory/src/test/java/com/google/sps/images/VisionManagerTest.java
jelares: Cloud vision API does not allow you to use the constructor for almost any of these internal objects - I am not sure why. Even for EntityAnnotation, I cannot find a way to construct it easily after a lot of searching. Do you have any possible ideas?

backstory/src/test/java/com/google/sps/images/VisionManagerTest.java
jelares: I don't think we have to, because every remaining line of code in the method are method calls from Vision API, which are already tested.

backstory/src/main/java/com/google/sps/images/VisionManager.java
jelares: Within the method itself, if responses.size() does not equal one, then an exception is thrown.

backstory/src/main/java/com/google/sps/images/VisionManager.java
ykjchen: Seems like you could do this then?

```
this(ImageAnnotatorClient.create());
```

backstory/src/test/java/com/google/sps/images/VisionManagerTest.java
ykjchen: List can be real (new ArrayList);

For the other types, builders should be exposed.

AnnotateImageResponse response = new AnnotateImageResponse.Builder().labelAnnotationsList(labelAnnotationsList).build();

backstory/src/test/java/com/google/sps/images/VisionManagerTest.java
ykjchen: But it seems like the manager's handling of the response is untested then?

backstory/src/main/java/com/google/sps/images/VisionImagesManager.java
```diff
@@ -0,0 +1,106 @@
...
+ */
+public final class VisionImagesManager implements ImagesManager {
+  private final ImageAnnotatorClient imageAnnotatorClient;
+
+  /**
+   * Creates a vision manager object using the image annotator client from Vision API.
+   */
+  public VisionImagesManager() throws IOException {
+    ImageAnnotatorClient imageAnnotatorClient = ImageAnnotatorClient.create();
+    this.imageAnnotatorClient = imageAnnotatorClient;
```
ykjchen, line 45: You should be able to do this(ImageAnnotatorClient.create());

backstory/src/main/java/com/google/sps/images/VisionImagesManager.java
```diff
@@ -0,0 +1,106 @@
...
+    AnnotateImageRequest request =
+        AnnotateImageRequest.newBuilder().addFeatures(feat).setImage(img).build();
+    requests.add(request);
+
+    // The invocation of batchAnnotateImages() makes a network call.
+    BatchAnnotateImagesResponse response = imageAnnotatorClient.batchAnnotateImages(requests);
+    List<AnnotateImageResponse> responses = response.getResponsesList();
+
+    // TODO: parallelize calls to detectLabelsFromImageBytes() since it requires a network call.
+    if (!(responses.size() == 1)) {
```
ykjchen, line 92: Easier to read would be `if (response.size() != 1) {`

PR #148, opened 2020-08-10
8/10: Make files pretty

annemariecabs:
Run `make pretty` and `make validate` on all files for Perspective project. Upload updated files.

Note: needed because I forgot to run `make pretty` on some files.


ykjchen:
Hmmm seems like the workflow check is not completely working, if previous builds did not fail. Created https://github.com/googleinterns/step174-2020/issues/151 to investigate.

PR #156, opened 2020-08-13
8/12: Update front end for Vision Demo

annemariecabs:
Update style.css, vision-demo.html, and vision-demo.js so that the front end of the vision demo matches the rest of website. Standardize rest of front end with any changes made in this update. Re-add demo files that were accidentally deleted in MVP commit.


backstory/src/demo/java/com/google/sps/servlets/AnalyzeImageServlet.java
```diff
@@ -0,0 +1,173 @@
+// Copyright 2020 Google LLC
```
jelares, line 1: If we include these files here, I can update them in a new PR to be functional with the front-end.

backstory/src/demo/java/com/google/sps/servlets/GetAnalyzedImagesServlet.java
```diff
@@ -0,0 +1,60 @@
+// Copyright 2020 Google LLC
```
jelares, line 1: If we include these files here, I can update them in a new PR to be functional with the front-end.

backstory/src/main/webapp/style.css
```diff
@@ -278,6 +277,68 @@ textarea {
...
+  width: 80%;
+}
+
+#image-upload-form {
+  text-align: center;
+  width: 100%;
+}
+
+/** Make the image upload input invisible */
+.file-input {
```
ykjchen, line 292: Looks like this was already declared above.

backstory/src/demo/webapp/vision-demo.js
```diff
@@ -0,0 +1,163 @@
...
+
+    const labelData = row.insertCell(0);
+    labelData.innerText = label.description;
+    labelData.classList.add('labels-table-cell');
+
+    const scoreData = row.insertCell(1);
+    scoreData.innerText = `${(label.score * 100).toFixed(2)}%`;
+    scoreData.classList.add('labels-table-cell');
+
+    index++;
```
ykjchen, line 139: Seems like if this was moved to the first line of this for loop, you could avoid the additional statement on line 124 above.

backstory/src/demo/webapp/vision-demo.js
```diff
@@ -0,0 +1,163 @@
...
+ * Format labels as an HTML table.
+ * 
+ * @param {object} labels - an array of labels (as JSON)
+ * @return {object} - a table representing the labels array
+ */
+function formatLabelsAsTable(labels) {
+  const table = document.createElement('table');
+  table.classList.add('labels-table');
+
+  let index = 0;
```
ykjchen, line 110: `index` works within a small scope like a for loop (it's clear that it's the incrementer index), but at this scope I'd suggest something more meaningful. Maybe `rowNumber`?

backstory/src/demo/java/com/google/sps/servlets/AnalyzeImageServlet.java
```diff
@@ -0,0 +1,173 @@
...
+      if (b.length < fetchSize) {
+        continueReading = false;
+      }
+
+      currentByteIndex += fetchSize;
+    }
+
+    return outputBytes.toByteArray();
+  }
+}
```
ykjchen, line 173: I think @jelares mentioned he may create utilities that would avoid the need to duplicate some of this code? Avoiding duplication is advantageous because it avoid the need to update code in multiple places when code changes are needed (also reduces risk of forgetting to update). 
See https://en.wikipedia.org/wiki/Duplicate_code

backstory/src/main/webapp/style.css
annemariecabs: Will remove this. Thank you!

annemariecabs: Done

backstory/src/demo/webapp/vision-demo.js
annemariecabs: That makes sense. I'll make the change.

backstory/src/demo/webapp/vision-demo.js
annemariecabs: That would definitely clarify this variable's purpose. I'll rename it to `rowNumber`.

backstory/src/demo/java/com/google/sps/servlets/AnalyzeImageServlet.java
annemariecabs: I think he is going to rework the code that appears in here. I'm happy to take it out if need be. I didn't code any of these servlet classes, they just appeared in this PR because they were deleted in an earlier PR. Would you rather I take this out or leave it in so @jelares can rework it using these classes?

backstory/src/demo/webapp/vision-demo.js
annemariecabs: Done

backstory/src/demo/webapp/vision-demo.js
annemariecabs: Done

backstory/src/demo/webapp/vision-demo.js
```diff
@@ -0,0 +1,162 @@
...
+ */
+function updateFileName() {
+  const fileInput = document.getElementById('image-upload');
+  const label = document.getElementById('upload-visual');
+
+  if (fileInput.files) {
+    if (fileInput.files.length > 1)  {
+      label.innerText = `${fileInput.files.length} files selected`;
+    }
+    else {
```
ykjchen, line 158: Google style uses K&R for braces:
https://google.github.io/styleguide/jsguide.html#formatting-nonempty-blocks

backstory/src/demo/webapp/vision-demo.js
```diff
@@ -0,0 +1,162 @@
...
+ * Updates the text of the file upload label to match the uploaded file
+ * or {number of files uploaded} files selected, if multiple files.
+ */
+function updateFileName() {
+  const fileInput = document.getElementById('image-upload');
+  const label = document.getElementById('upload-visual');
+
+  if (fileInput.files) {
+    if (fileInput.files.length > 1)  {
+      label.innerText = `${fileInput.files.length} files selected`;
```
ykjchen, line 156: Nit: assigning fileInput.files.length to a local variable avoids the need to access the property twice.

backstory/src/demo/java/com/google/sps/servlets/AnalyzeImageServlet.java
```diff
@@ -0,0 +1,173 @@
...
+
+    if (bytes == null || imageUrl == null) {
+      // Redirect back to the HTML page.
+      response.sendError(400, "Please upload a valid image.");
+
+    } else {
+      // Gets the full label information from the image byte array by calling Images Manager.
+      ImagesManager demoImagesManager = new VisionImagesManager();
+      List<byte[]> imagesAsByteArrays = new ArrayList<>();
+      imagesAsByteArrays.add(bytes);
```
ykjchen, line 76: Can the list be initialized with `bytes` already in it, rather than initializing an empty list, then adding `bytes`?

backstory/src/demo/java/com/google/sps/servlets/AnalyzeImageServlet.java
```diff
@@ -0,0 +1,173 @@
...
+    // Get the input from the form.
+    // Get the URL of the image that the user uploaded to Blobstore.
+    final String imageUrl = getUploadedFileUrl(request, "image-upload-form");
+
+    // Get the raw byte array representing the image from Blobstore
+    final byte[] bytes = getBlobBytes(request, "image-upload-form");
+
+    if (bytes == null || imageUrl == null) {
+      // Redirect back to the HTML page.
+      response.sendError(400, "Please upload a valid image.");
```
ykjchen, line 70: It would be clearer here to have an early return.
1. Without the early return I had to scoll to the bottom of this method to make sure no other code runs after this.
2. The early return reduces nesting (improves readability) of the block below:

```
if (...) {
  response.sendError(400, "Please upload a valid image.");
  return;
}  // Note: no else block (to reduce nesting).

ImagesManager demoImagesManager = new VisionImagesManager();
...
```

backstory/src/demo/java/com/google/sps/servlets/AnalyzeImageServlet.java
```diff
@@ -0,0 +1,173 @@
...
+    }
+
+    ByteArrayOutputStream outputBytes = new ByteArrayOutputStream();
+
+    int fetchSize = BlobstoreService.MAX_BLOB_FETCH_SIZE;
+    long currentByteIndex = 0;
+    boolean continueReading = true;
+    while (continueReading) {
+      // end index is inclusive, so we have to subtract 1 to get fetchSize bytes
+      byte[] b =
```
ykjchen, line 159: Can you name this a full word? I'm not sure what `b` is supposed to stand for.

backstory/src/demo/java/com/google/sps/servlets/AnalyzeImageServlet.java
ykjchen: So it sounds like you plan to check this in here, and then @jelares will refactor this code into a shared class later? We'll just need to make sure that refactoring happens before the Backstory servlet PR, to avoid too much code duplication.

annemariecabs: That makes sense. I believe this is what Jesus intends to do. I'll merge this now, and send a heads up to @jelares that this is the plan.

annemariecabs: Done

backstory/src/demo/webapp/vision-demo.js
annemariecabs: Thanks! Will change this.

backstory/src/demo/webapp/vision-demo.js
annemariecabs: I agree. I'll make the change.

backstory/src/demo/java/com/google/sps/servlets/AnalyzeImageServlet.java
annemariecabs: Unfortunately, I don't think it can be initialized that way other than with Arrays.asList() but that could be risky. I didn't actually write this code (it's Jesus's), so I'm not sure.

backstory/src/demo/java/com/google/sps/servlets/AnalyzeImageServlet.java
annemariecabs: I'll make this change.

backstory/src/demo/java/com/google/sps/servlets/AnalyzeImageServlet.java
annemariecabs: I renamed this to blobstoreData temporarily to fix this issue. I'm not sure if this is the best name though so @jelares if you want to change it, please do.

backstory/src/demo/webapp/vision-demo.js
annemariecabs: Done

backstory/src/demo/webapp/vision-demo.js
annemariecabs: Done

PR #161, opened 2020-08-15
8/14: Add more metrics to Content Decision

annemariecabs:
Add checks for obscenity, profanity, sexually explicit material, and offensive material into the ContentDecisions class in the logic for deciding if stories are appropriate (this should allow us to more effectively filter out non-family friendly material). Add tests to ensure these new metrics are filtering out content.

Note: thresholds set to filter content in this PR were decided based off experimenting with the Perspective API.

### DIFF
--- a/backstory/src/main/java/com/google/sps/perspective/data/ContentDecisions.java
+++ b/backstory/src/main/java/com/google/sps/perspective/data/ContentDecisions.java
@@ -22,6 +22,21 @@ import java.util.Map;
  * is appropriate using analysis from Perspective API.
  */
 public class ContentDecisions {
+  
+  // threshold here means that score must be below this threshold 
+  // in order to be considered appropriate.
+
+  /** the threshold for appropriateness for toxicity score */
+  public static final float TOXICITY_THRESHOLD = .7f;
+  /** the threshold for appropriateness for sexually explicit score */
+  public static final float SEXUALLY_EXPLICIT_THRESHOLD = .6f;
+  /** the threshold for appropriateness for profanity score */
+  public static final float PROFANITY_THRESHOLD = .8f;
+  /** the threshold for appropriateness for offensiveness (identity attack) score */
+  public static final float OFFENSIVE_THRESHOLD = .8f;
+  /** the threshold for appropriateness for obscenity score */
+  public static final float OBSCENITY_THRESHOLD = .8f;
+
   /**
    * Overrides default constructor to ensure class can't be instantiated.
    */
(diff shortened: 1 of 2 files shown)


backstory/src/main/java/com/google/sps/perspective/data/ContentDecisions.java
```diff
@@ -69,4 +71,104 @@ private static boolean isToxic(Map<AttributeType, Float> attributeTypesToScores)
...
+   *     identity attack score
+   */
+  private static boolean isOffensive(Map<AttributeType, Float> attributeTypesToScores)
+      throws IllegalArgumentException {
+    if (attributeTypesToScores == null) {
+      throw new IllegalArgumentException("Map (attributeTypesToScores) cannot be null.");
+    } else if (!attributeTypesToScores.containsKey(AttributeType.IDENTITY_ATTACK)) {
+      throw new IllegalArgumentException(
+          "Map (attributeTypesToScores) does not contain an identity attack score");
+    }
```
ykjchen, line 143: There's some duplication here... this may be easier to maintain if moved to a helper method:

```
void validateHasScoreForAttributeType(Map<AttributeType, Float> attributeTypesToScores, AttributeType attributeType) throws IllegalArgumentException;
```

backstory/src/main/java/com/google/sps/perspective/data/ContentDecisions.java
```diff
@@ -69,4 +71,104 @@ private static boolean isToxic(Map<AttributeType, Float> attributeTypesToScores)
...
+    if (attributeTypesToScores == null) {
+      throw new IllegalArgumentException("Map (attributeTypesToScores) cannot be null.");
+    } else if (!attributeTypesToScores.containsKey(AttributeType.SEXUALLY_EXPLICIT)) {
+      throw new IllegalArgumentException(
+          "Map (attributeTypesToScores) does not contain a sexually explicit score");
+    }
+
+    float sexuallyExplicit = attributeTypesToScores.get(AttributeType.SEXUALLY_EXPLICIT);
+
+    return sexuallyExplicit >= .6f;
```
ykjchen, line 97: Maybe listing these thresholds as constants at the top of the file would make them easier to find and change.

Making them public would then allow you to reference the constants from the tests, which could reduce test maintenance:

```
void checkDecisionForHighSexuallyExplicit() {
  final float HIGH_SEXUALLY_EXPLICIT = ContentDecisions.MAX_APPROPRIATE_SEXUALLY_EXPLICIT_SCORE + 0.1f;
 ...
}
```

backstory/src/main/java/com/google/sps/perspective/data/ContentDecisions.java
annemariecabs: i can do that. Thank you for the suggestion!

backstory/src/main/java/com/google/sps/perspective/data/ContentDecisions.java
annemariecabs: This is a great suggestion. I'll implement this.

backstory/src/main/java/com/google/sps/perspective/data/ContentDecisions.java
annemariecabs: Done

backstory/src/main/java/com/google/sps/perspective/data/ContentDecisions.java
annemariecabs: Done

PR #162, opened 2020-08-15
8/14: Fix bug in Perspective Demo

annemariecabs:
Adapt PerspectiveDecision and the generatePerspectiveDecision method of PerspectiveStoryAnalysisManager to work with needs of the PerspectiveAnalysisManager and avoid NullPointerException (which occurs currently when a not appropriate story is analyzed in demo). Update PerspectiveServlet with this information.

Note: the bug was that in the demo, when a story was judged not appropraite, a NullPointerError was thrown when trying to access its values through the PerspectiveDecision because what actually was returned was not a PerspectiveDecision but was null. This class changed the PerspectiveDecision in order to avoid this.

### DIFF
--- a/backstory/src/main/java/com/google/sps/perspective/PerspectiveStoryAnalysisManager.java
+++ b/backstory/src/main/java/com/google/sps/perspective/PerspectiveStoryAnalysisManager.java
@@ -110,22 +110,14 @@ public class PerspectiveStoryAnalysisManager implements StoryAnalysisManager {
    * as a PerspectiveDecision object (for demo purposes).
    *
    * @param story The story to be analyzed
-   * @return An object describing the recommendation resulting from the analysis.
-   * @throws NoAppropriateStoryException if story is not considered appropriate
+   * @return An object describing the recommendation resulting from the analysis. If the
+   *    the PerspectiveDecision story field is null, it's not an appropriate story.
    */
-  public PerspectiveDecision generatePerspectiveDecision(String story)
-      throws NoAppropriateStoryException {
+  public PerspectiveDecision generatePerspectiveDecision(String story) {
     PerspectiveAPIClient apiClient = new PerspectiveAPIClient(perspectiveAPI);
     PerspectiveValues storyValues = apiClient.analyze(Arrays.asList(REQUESTED_ATTRIBUTES), story);
     boolean isStoryAppropriate = ContentDecisions.makeDecision(storyValues);
 
-    // if content decisions returns that it's appropriate
-    // then return a StoryDecision object with this story
-    if (isStoryAppropriate) {
-      return new PerspectiveDecision(story, storyValues);
-    }
-
-    // otherwise throw the NoAppropriateStoryException
-    throw new NoAppropriateStoryException("The story passed in was not appropriate.");
+    return new PerspectiveDecision(story, isStoryAppropriate, storyValues);
   }
 }
--- a/backstory/src/demo/java/com/google/sps/servlets/PerspectiveServlet.java
+++ b/backstory/src/demo/java/com/google/sps/servlets/PerspectiveServlet.java
@@ -79,15 +79,9 @@ public final class PerspectiveServlet extends HttpServlet {
     }
 
     // generate a PerspectiveDecision with that manager
-    PerspectiveDecision perspectiveDecision;
-    try {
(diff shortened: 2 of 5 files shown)


backstory/src/main/java/com/google/sps/perspective/PerspectiveStoryAnalysisManager.java
```diff
@@ -110,22 +110,14 @@ public StoryDecision generateDecision(String story) throws NoAppropriateStoryExc
...
     boolean isStoryAppropriate = ContentDecisions.makeDecision(storyValues);
 
-    // if content decisions returns that it's appropriate
-    // then return a StoryDecision object with this story
-    if (isStoryAppropriate) {
-      return new PerspectiveDecision(story, storyValues);
-    }
-
-    // otherwise throw the NoAppropriateStoryException
-    throw new NoAppropriateStoryException("The story passed in was not appropriate.");
```
ykjchen, line 129: Can the NoAppropriateStoryException file be deleted too?

backstory/src/main/java/com/google/sps/perspective/PerspectiveStoryAnalysisManager.java
```diff
@@ -110,22 +110,14 @@ public StoryDecision generateDecision(String story) throws NoAppropriateStoryExc
...
+    return new PerspectiveDecision(story, isStoryAppropriate, storyValues);
   }
```
ykjchen, line 122: No tests needed to be updated?

backstory/src/main/java/com/google/sps/perspective/PerspectiveStoryAnalysisManager.java
annemariecabs: This is only updated for PerspectiveDecision, which is the version of StoryDecision that the Perspective demo uses (but NOT the BackstoryServlet) as such NoAppropriateStoryException is still necessary because it's thrown in StoryDecision.

backstory/src/main/java/com/google/sps/perspective/PerspectiveStoryAnalysisManager.java
annemariecabs: This is only changing elements used in the demo, which we decided not to test.

backstory/src/main/java/com/google/sps/perspective/PerspectiveStoryAnalysisManager.java
annemariecabs: Done

backstory/src/main/java/com/google/sps/perspective/PerspectiveStoryAnalysisManager.java
annemariecabs: Done

ykjchen: I think that's fine for now... but I should probably clarify that when deciding what to test, it would be safer to make that distinction based on where the code lives. If it lives in /demo then it's safe to avoid testing, if it lives in /src/main, then there's nothing stopping people from using this method once it's added to the public API. In general we consider the public API part of the true functionality of a class, even if currently it is exposed just for tests/demos.

backstory/src/main/java/com/google/sps/perspective/PerspectiveStoryAnalysisManager.java
ykjchen: Thanks for explaining.

backstory/src/main/java/com/google/sps/perspective/PerspectiveStoryAnalysisManager.java
annemariecabs: Ok. That makes sense. I'll keep that in mind for the future.

annemariecabs:
> @annemariecabs could the PR title include specifics about what bug was fixed? That would help make github commit history clearer.

I added info on what the bug was in the description.

backstory/src/main/java/com/google/sps/perspective/PerspectiveStoryAnalysisManager.java
annemariecabs: Done

PR #163, opened 2020-08-17
8/17: Add StoryEndingTools class and corresponding tests

annemariecabs:
Add StoryEndingTools class, which is a static library class, whose most important method (endStory) removes sentence fragments at the end of stories and appends a story ending in order to make the endings of stories seem more natural. (This is necessary because GPT-2 has no natural ending programmed into it, which means that it can just end in the middle of a sentence). Adds corresponding unit testing.


backstory/src/main/java/com/google/sps/story/data/StoryEndingTools.java
```diff
@@ -0,0 +1,106 @@
...
+    }
+
+    int lastPeriodIndex = story.lastIndexOf(".");
+    int lastExclamationPointIndex = story.lastIndexOf("!");
+    int lastQuestionMarkIndex = story.lastIndexOf("?");
+
+    // find the last sentence-ending punctuation by finding the max number
+    // amongst all the sentence-ending types indices
+    int lastSentenceEnder = Math.max(lastPeriodIndex, lastExclamationPointIndex);
+    lastSentenceEnder = Math.max(lastSentenceEnder, lastQuestionMarkIndex);
```
ykjchen, line 80: Consider making this scalable by storing the endings in an array and iterating through them.

backstory/src/main/java/com/google/sps/story/data/StoryEndingTools.java
```diff
@@ -0,0 +1,106 @@
...
+   * or question mark).
+   *
+   * @param story to remove the sentence fragment from
+   * @return the passed-in story without the last sentence fragment
+   * @throws IllegalArgumentException if story is null
+   */
+  public static String removeSentenceFragmentAtEnd(String story) {
+    if (story == null) {
+      throw new IllegalArgumentException("Story should not be null");
+    }
```
ykjchen, line 71: Also check length of story after trimming whitespace?

backstory/src/main/java/com/google/sps/story/data/StoryEndingTools.java
```diff
@@ -0,0 +1,106 @@
...
+   * the story with that ending.
+   *
+   * @param story the story to add an ending to
+   * @return story + ending
+   * @throws IllegalArgumentException if story is null
+   */
+  public static String addEnding(String story) {
+    if (story == null) {
+      throw new IllegalArgumentException("Story should not be null");
+    }
```
ykjchen, line 102: Also check length of story after trimming whitespace?

backstory/src/test/java/com/google/sps/story/StoryEndingToolsTest.java
```diff
@@ -0,0 +1,209 @@
...
+    // be same length as it was before
+    for (String sentence: COMPLETE_SENTENCES) {
+      String output = StoryEndingTools.endStory(sentence);
+      // remove last sentence by removing punctuation then calling remove fragment
+      String outputWithoutEnding = StoryEndingTools.removeSentenceFragmentAtEnd(output.substring(0, output.length() - 1));
+
+      Assert.assertEquals(sentence.length(), outputWithoutEnding.length());
+    }
+  }
+}
```
ykjchen, line 209: Is it possible to reduce the amount of logic in these tests?

See: https://drive.google.com/file/d/1QbOCmyI7hwhCUlHJ6NeH0Jfk7oqxZoPB/view

backstory/src/main/java/com/google/sps/story/data/StoryEndingTools.java
```diff
@@ -0,0 +1,106 @@
...
+   * Takes a story from GPT-2 and returns onee with a more natural ending.
+   *
+   * @param story the story to return with a natural ending
+   * @return story with ending
+   * @throws IllegalArgumentException if story is null
+   */
+  public static String endStory(String story) {
+    if (story == null) {
+      throw new IllegalArgumentException("Story should not be null");
+    }
```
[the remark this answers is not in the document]
annemariecabs: I'm happy to make a change, but a bit confused. I don't believe I trim whitespace in this file. What is "trimming whitespace"  referring to in this comment? @ykjchen

backstory/src/main/java/com/google/sps/story/data/StoryEndingTools.java
annemariecabs: Duplicate comment - refer to one above.

backstory/src/main/java/com/google/sps/story/data/StoryEndingTools.java
annemariecabs: Similarly to above, confused as to what trimming whitespace refers to.

backstory/src/main/java/com/google/sps/story/data/StoryEndingTools.java
annemariecabs: I can make that change. Thanks!

annemariecabs: Done

backstory/src/test/java/com/google/sps/story/StoryEndingToolsTest.java
annemariecabs: It is possible to reduce the amount of logic. I see what they mean about it being more about input-output pairs. As such, I think that I'm going to rewrite emptyInputForRemoveFragment, fragmentInputForRemoveFragment, and sentenceInputWithRemoveFragment to be more like this as well as ensureFragmentRemoved. For the others, I believe the logic is necessary because it's more difficult to test those (as they get and attach a random ending), and I think the logic is fairly simple and allows testing on more inputs.

annemariecabs: Actually, in an **update**, I changed all of the methods, and there should be substantially less logic, and hopefully all of it is simpler.

annemariecabs: Done

backstory/src/main/java/com/google/sps/story/data/StoryEndingTools.java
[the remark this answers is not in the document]
ykjchen: Like this:
```
if (story == null || story.trim().length() == 0) { ...
```

This would ensure that this is not just an empty string.

https://www.geeksforgeeks.org/java-string-trim-method-example/#:~:text=trim()is%20a%20built,and%20returns%20the%20omitted%20string.

backstory/src/main/java/com/google/sps/story/data/StoryEndingTools.java
ykjchen: Since this type of validation is repeated, consider adding a helper method so that updates to the validation don't need to be duplicated.

- (void)validateStory(String story) throws IllegalArgumentException;

backstory/src/main/java/com/google/sps/story/data/StoryEndingTools.java
```diff
@@ -0,0 +1,111 @@
...
+   * @return the passed-in story without the last sentence fragment
+   * @throws IllegalArgumentException if story is null
+   */
+  public static String removeSentenceFragmentAtEnd(String story) {
+    if (story == null) {
+      throw new IllegalArgumentException("Story should not be null");
+    }
+
+    // find the last sentence-ending punctuation
+    int lastSentenceEnder = -1;
```
ykjchen, line 77: Consider:

```
final int NOT_FOUND = -1;
int lastSentenceEnder = NOT_FOUND;
for ...;
if (lastSentenceEnder == NOT_FOUND) {
  throw ...;
}
```

Or else you could define what should happen if the story passed in is itself one large fragment. Should the return value in that case just be an ending?

backstory/src/main/java/com/google/sps/story/data/StoryEndingTools.java
```diff
@@ -0,0 +1,111 @@
...
+    for (String ender : SENTENCE_ENDERS) {
+      int lastIndex = story.lastIndexOf(ender);
+
+      if (lastIndex > lastSentenceEnder) {
+        lastSentenceEnder = lastIndex;
+      }
+    }
+
+    // if last period coincides with end of story, then return story as is
+    if (lastSentenceEnder == story.length() - 1) {
```
ykjchen, line 88: It looks like this check relies on `lastSentenceEnder` being initialized to -1. This is subtle and easy to miss during maintenance, and could be the cause of bugs. It would be safer to first check that an ender was indeed found before getting to this point.

backstory/src/test/java/com/google/sps/story/StoryEndingToolsTest.java
```diff
@@ -0,0 +1,229 @@
...
+  private final static String EXCLAMATION_FRAGMENT = "Oh no! This isn't ";
+  private final static String COMPLETE_EXCLAMATION = "I had no idea. Oh no!";
+
+  // aggregate all of these into one input array
+  private final static String[] INPUTS = { EMPTY_TEXT, SENTENCE_FRAGMENT, COMPLETE_SENTENCE, 
+                                    QUESTION_FRAGMENT, COMPLETE_QUESTION, 
+                                    EXCLAMATION_FRAGMENT, COMPLETE_EXCLAMATION };
+
+  // array of sentence enders
+  private final static String[] SENTENCE_ENDERS = {".", "?", "!"};
```
ykjchen, line 50: Can you make the StoryEndingTools.SENTENCE_ENDERS field public instead? That avoids the need to keep these two in sync.

backstory/src/test/java/com/google/sps/story/StoryEndingToolsTest.java
```diff
@@ -0,0 +1,229 @@
...
+  /**
+   * Private helper method to check if text ends in punctuation.
+   * 
+   * @param text the text to check ending for
+   * @return true, if it's empty or if the char at the last index is in SENTENCE_ENDERS
+   *
+   */
+  private static boolean endsInSentenceEndingPunctuation(String text) {
+    if (text.equals("")) {
+      return true;
```
ykjchen, line 89: This adds an element of uncertainty to the tests:

```
String sentence = foo();
Assert.assertTrue(endsInSentenceEndingPunctuation(sentence));
```

If I see this, I won't be certain that sentence indeed has punctuation.

I'd recommend either failing the test here, or removing this block, to avoid introducing additional complexity.

backstory/src/main/java/com/google/sps/story/data/StoryEndingTools.java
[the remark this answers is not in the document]
annemariecabs: That makes sense. Will do.

backstory/src/main/java/com/google/sps/story/data/StoryEndingTools.java
annemariecabs: That makes sense. I'll add this.

backstory/src/main/java/com/google/sps/story/data/StoryEndingTools.java
annemariecabs: I'll make the change you suggested above here as well.

backstory/src/main/java/com/google/sps/story/data/StoryEndingTools.java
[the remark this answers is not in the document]
annemariecabs: Done

backstory/src/main/java/com/google/sps/story/data/StoryEndingTools.java
annemariecabs: Done

backstory/src/main/java/com/google/sps/story/data/StoryEndingTools.java
annemariecabs: I think that the story in that case should be just an ending, so I'm going to not throw an exception.

backstory/src/main/java/com/google/sps/story/data/StoryEndingTools.java
annemariecabs: Done

backstory/src/main/java/com/google/sps/story/data/StoryEndingTools.java
annemariecabs: I can add this constant, but I don't think the exception throw is appropriate because fragments passed-in should just return an ending.

backstory/src/main/java/com/google/sps/story/data/StoryEndingTools.java
annemariecabs: That makes sense. I'll add a check that the ender was found by checking it's not equal to `NOT_FOUND` before executing this.

backstory/src/test/java/com/google/sps/story/StoryEndingToolsTest.java
annemariecabs: Sure.

backstory/src/test/java/com/google/sps/story/StoryEndingToolsTest.java
annemariecabs: I'll remove this block because I agree, it isn't relevant.

annemariecabs: Done

backstory/src/test/java/com/google/sps/story/StoryEndingToolsTest.java
annemariecabs: Done

backstory/src/main/java/com/google/sps/story/data/StoryEndingTools.java
annemariecabs: Done

backstory/src/main/java/com/google/sps/story/data/StoryEndingTools.java
annemariecabs: Done

backstory/src/main/java/com/google/sps/story/data/StoryEndingTools.java
```diff
@@ -0,0 +1,119 @@
...
+  /**
+   * Remove a sentence fragment at the end of the story (if there is one)
+   * as that commonly occurs with GPT2-generated text. Returns the story
+   * up to the last sentence-ending punctuation (period, exclamation point,
+   * or question mark).
+   *
+   * @param story to remove the sentence fragment from
+   * @return the passed-in story without the last sentence fragment
+   * @throws IllegalArgumentException if story is null
+   */
```
jelares, line 68: Great documentation.

annemariecabs: Thanks!

PR #165, opened 2020-08-18
8/18: Add Datamuse Client and integrate with project

annemariecabs:
Add DatamuseRequestClient, which allows for requests to be made of the Datamuse API to allow for adjective generation based on a noun. Move APINotAvailableException so it can be used in this class and update import statements accordingly. Add testing in DatamuseRequestClientTest and add HttpUrlStreamHandler to enable testing with classes that use URLs (specifically DatamuseRequestClient).


backstory/src/main/java/com/google/sps/story/DatamuseRequestClient.java
```diff
@@ -0,0 +1,150 @@
...
+  /**
+   * Constructs a request client with the default Datamuse url.
+   */
+  public DatamuseRequestClient() {
+    this(DATAMUSE_URL);
+  }
+
+  /**
+   * Constructs a request client that will query
+   * the given URL (for testing purposes)
```
ykjchen, line 47: It's typically discouraged to expose functionality "only for testing". Once a method is public, we should expect that it may also be used by real code. Suggest removing the "for testing purposes" from the docs, and just consider this part of the class's public API.

backstory/src/main/java/com/google/sps/story/DatamuseRequestClient.java
```diff
@@ -0,0 +1,150 @@
...
+   * @param cap the maximum number of adjectives to retrieve
+   * @return an array of adjectives related to the noun
+   * @throws IllegalArgumentException if noun is more than one word (has a space in it)
+   * @throws APINotAvailableException if Datamuse API cannot be reached
+   * @throws RuntimeException if JSON received back from the API cannot be parsed (JSONException)
+   *    or does not have objects of type JSON (ClassCastException)
+   */
+  public String[] getRelatedAdjectives(String noun, int cap)
+      throws IllegalArgumentException, APINotAvailableException, RuntimeException {
+    if (noun.contains(" ")) {
```
ykjchen, line 71: Do you want to also check for other whitespace characters?
https://stackoverflow.com/a/4067825

backstory/src/main/java/com/google/sps/story/DatamuseRequestClient.java
```diff
@@ -0,0 +1,150 @@
...
+   *
+   * @param noun the noun to get adjectives related to (must be one word)
+   * @param cap the maximum number of adjectives to retrieve
+   * @return an array of adjectives related to the noun
+   * @throws IllegalArgumentException if noun is more than one word (has a space in it)
+   * @throws APINotAvailableException if Datamuse API cannot be reached
+   * @throws RuntimeException if JSON received back from the API cannot be parsed (JSONException)
+   *    or does not have objects of type JSON (ClassCastException)
+   */
+  public String[] getRelatedAdjectives(String noun, int cap)
```
ykjchen, line 69: I'd suggest naming this "fetch" rather than "get", since "get" can sometimes be associated with a fast property getter.

backstory/src/main/java/com/google/sps/story/DatamuseRequestClient.java
```diff
@@ -0,0 +1,150 @@
...
+   * Query the site at the given url and retrieve the content (likely JSON) from that
+   * site. If there's nothing at that site, return null. This code was
+   * adapted from the DatamuseQuery class in the Datamuse4J repo
+   * (https://github.com/sjblair/Datamuse4J/blob/master/src/datamuse/DatamuseQuery.java).
+   *
+   * @param url the url to query
+   * @return the content at the site.
+   * @throws APINotAvailableException if the method can't retrieve the content at the site
+   */
+  private String querySite(String url) throws APINotAvailableException {
```
ykjchen, line 120: I was a bit confused about what this method does before reading the implementation. Could this be named `queryUrl()`? As far as I know "site" is not commonly used to refer to a URL or resource.

backstory/src/main/java/com/google/sps/story/DatamuseRequestClient.java
```diff
@@ -0,0 +1,150 @@
...
+      throw new RuntimeException("JSON array received was not of JSON objects.", exception);
+    } catch (JSONException exception) {
+      throw new RuntimeException(
+          "JSON Object did not have a String value for the key \"word\".", exception);
+    }
+  }
+
+  /**
+   * Query the site at the given url and retrieve the content (likely JSON) from that
+   * site. If there's nothing at that site, return null. This code was
```
ykjchen, line 112: Suggest removing "site" from the docs, as I think that's often used to represent something more than a HTTP resource.

I'd also recommend stating what type of request is being made, e.g. `Makes a GET request to the provided url.`

backstory/src/main/java/com/google/sps/story/DatamuseRequestClient.java
```diff
@@ -0,0 +1,150 @@
...
+   * adapted from the DatamuseQuery class in the Datamuse4J repo
+   * (https://github.com/sjblair/Datamuse4J/blob/master/src/datamuse/DatamuseQuery.java).
+   *
+   * @param url the url to query
+   * @return the content at the site.
```
ykjchen, line 117: Is this more specifically "the response body of the GET request"?

backstory/src/main/java/com/google/sps/story/DatamuseRequestClient.java
```diff
@@ -0,0 +1,150 @@
...
+   * @throws APINotAvailableException if the method can't retrieve the content at the site
+   */
+  private String querySite(String url) throws APINotAvailableException {
+    final String ERROR_MESSAGE =
+        "This query could not successfully retrieve content from Datamuse API.\n";
```
ykjchen, line 122: Optional: using a platform independent newline token would improve portability:
https://www.baeldung.com/java-string-newline#2-using-platform-independent-line-separators

backstory/src/test/java/com/google/sps/story/DatamuseRequestClientTest.java
```diff
@@ -0,0 +1,339 @@
...
+  private static final String DEFAULT_URL = "http://sample-website.com/fake?";
+
+  /** the stream handlet to use for tests in this class (necessary to mock URLConnection) */
+  private static HttpUrlStreamHandler httpUrlStreamHandler;
+ 
+  /** 
+   * Set up the URLStreamHandlerFactory with the HTTPUrlStreamHandler 
+   * in order to mock URLConnections to properly test methods 
+   * that call outside URLs. Code comes from this tutorial: 
+   * https://claritysoftware.co.uk/mocking-javas-url-with-mockito/.
```
ykjchen, line 65: Nice work on the research!

backstory/src/test/java/com/google/sps/story/DatamuseRequestClientTest.java
```diff
@@ -0,0 +1,339 @@
...
+
+/** Quality tests for DatamuseRequestClient */
+@RunWith(JUnit4.class)
+public final class DatamuseRequestClientTest {
+
+  /** a default url to test with */
+  private static final String DEFAULT_URL = "http://sample-website.com/fake?";
+
+  /** the stream handlet to use for tests in this class (necessary to mock URLConnection) */
+  private static HttpUrlStreamHandler httpUrlStreamHandler;
```
ykjchen, line 59: Can this not be static, so that state is less likely to carry over from test to test? (having a clean slate for each test avoids the need to wonder whether these tests are order-dependent)

backstory/src/test/java/com/google/sps/story/DatamuseRequestClientTest.java
```diff
@@ -0,0 +1,339 @@
...
+  /**
+   * Reset the URLConnections of this stream handler before every method,
+   * so that URLConnections injected for other tests do not mess 
+   * with the current test. Code taken from this tutorial: 
+   * https://claritysoftware.co.uk/mocking-javas-url-with-mockito/.
+   */
+  @Before
+  public void reset() {
+    httpUrlStreamHandler.resetConnections();
+  }
```
ykjchen, line 86: Seems like this could be avoided by making `httpUrlStreamHandler` non-static.

backstory/src/test/java/com/google/sps/story/DatamuseRequestClientTest.java
```diff
@@ -0,0 +1,339 @@
...
+    final int DEFAULT_MAX = 5;
+
+    DatamuseRequestClient client = new DatamuseRequestClient(DEFAULT_URL);
+
+    try {
+      client.getRelatedAdjectives(MULTIWORD_NOUN, DEFAULT_MAX);
+    } catch (IllegalArgumentException exception) {
+      return; // test should pass if this exception is thrown
+    } catch (APINotAvailableException | RuntimeException exception) {
+      // we dont' want these to be thrown
```
ykjchen, line 104: I remember seeing this in some of your other tests:
https://www.baeldung.com/junit-assert-exception#junit-4

Is that possible here too? That would be more concise/expressive and avoid excess logic/documentation.

backstory/src/main/java/com/google/sps/story/DatamuseRequestClient.java
annemariecabs: That makes sense. I'll remove it.

annemariecabs: Done

backstory/src/main/java/com/google/sps/story/DatamuseRequestClient.java
annemariecabs: That makes sense. I'll change it to `fetch`.

annemariecabs: Done

backstory/src/main/java/com/google/sps/story/DatamuseRequestClient.java
annemariecabs: Sure. I'll make the change.

backstory/src/main/java/com/google/sps/story/DatamuseRequestClient.java
annemariecabs: That makes sense.

backstory/src/main/java/com/google/sps/story/DatamuseRequestClient.java
annemariecabs: It could be. As I mentioned above, I don't know if this is actually considered a GET request. I'm happy to make the change. Please let me know your thoughts after reading the last comment. @ykjchen

backstory/src/main/java/com/google/sps/story/DatamuseRequestClient.java
annemariecabs: I'll change this. Thank you!

annemariecabs: Done

backstory/src/test/java/com/google/sps/story/DatamuseRequestClientTest.java
annemariecabs: Thanks!

backstory/src/test/java/com/google/sps/story/DatamuseRequestClientTest.java
annemariecabs: Unfortunately, I did that at first, and it was warning me I couldn't call this method without catching the exception (i.e. I couldn't just let it be thrown to test if it is thrown. As such, I adopted this alternate approach.

annemariecabs: Done

backstory/src/test/java/com/google/sps/story/DatamuseRequestClientTest.java
annemariecabs: As you noticed below, it does have a clean slate for every test. Doing this saves the effort of instantiating this class before every single test, when all you need to do is reset the connections. I think just resetting connections efficiently accomplishes this. However, if you want me to change it, I'm happy to. @ykjchen

backstory/src/test/java/com/google/sps/story/DatamuseRequestClientTest.java
annemariecabs: Refer to above comment. Please let me know if you still want me to change this.

backstory/src/test/java/com/google/sps/story/DatamuseRequestClientTest.java
annemariecabs: Thank you!

backstory/src/main/java/com/google/sps/story/DatamuseRequestClient.java
annemariecabs: Done

backstory/src/main/java/com/google/sps/story/DatamuseRequestClient.java
annemariecabs: I'm not sure if it's necessary, but I'm happy to make the change just to be safe.

annemariecabs: Done

backstory/src/main/java/com/google/sps/story/DatamuseRequestClient.java
annemariecabs: Done

backstory/src/test/java/com/google/sps/story/DatamuseRequestClientTest.java
ykjchen: It looks like you could instead declare that the test method throws these exceptions:
https://stackoverflow.com/a/31423712

Throwing would cause the test to fail.

backstory/src/test/java/com/google/sps/story/DatamuseRequestClientTest.java
ykjchen: Yes, don't worry about saving this level of computation during tests. Confidence in correctness is much more important.

backstory/src/main/java/com/google/sps/story/DatamuseRequestClient.java
ykjchen: So I'm guessing that this UrlConnection is a HTTPUrlConnection, and that by default, the method is `GET`:
https://docs.oracle.com/javase/7/docs/api/java/net/HttpURLConnection.html#method

i.e. the UrlConnection object is making a GET request to that URL, and then exposing the response as a stream.

You could add logging to confirm.

backstory/src/test/java/com/google/sps/story/DatamuseRequestClientTest.java
ykjchen: Yeah, it's possible that there's other state that's carried over (if not now, then if HttpUrlStreamHandler adds new functionality in the future)... it would be safer to start each test with as clean a slate as possible.

backstory/src/main/java/com/google/sps/story/DatamuseRequestClient.java
annemariecabs: In that case, I'm going to add that it was a GET request to the documentation.

annemariecabs: Done

backstory/src/main/java/com/google/sps/story/DatamuseRequestClient.java
annemariecabs: Yes, this seems like it is. Thank you!

annemariecabs: Done

backstory/src/test/java/com/google/sps/story/DatamuseRequestClientTest.java
annemariecabs: Ok I'll change it.

annemariecabs: Done

backstory/src/test/java/com/google/sps/story/DatamuseRequestClientTest.java
annemariecabs: I do agree, but unfortunately it looks like this choice is the only option because httpUrlStreamHandler seems like it can only be set once in the class. Please let me know if you see another option.

annemariecabs: Done

backstory/src/test/java/com/google/sps/story/DatamuseRequestClientTest.java
annemariecabs: I used the technique there to remove the catch statement for the other exceptions and kept the main catch statement related to the relevant exception so I could verify that that specific exception had been thrown. Hope this fixes it!

annemariecabs: Done

backstory/src/test/java/com/google/sps/story/DatamuseRequestClientTest.java
```diff
@@ -0,0 +1,337 @@
...
+  public void multipleWordNoun() throws Exception {
+    final String MULTIWORD_NOUN = "multiword noun";
+    final int DEFAULT_MAX = 5;
+
+    DatamuseRequestClient client = new DatamuseRequestClient(DEFAULT_URL);
+
+    try {
+      client.fetchRelatedAdjectives(MULTIWORD_NOUN, DEFAULT_MAX);
+    } catch (IllegalArgumentException exception) {
+      return; // test should pass if this exception is thrown
```
ykjchen, line 102: Can you use the `expected` annotation instead (doesn't marking this method with `throws` make this possible)?

Question also applies to tests below.

backstory/src/test/java/com/google/sps/story/DatamuseRequestClientTest.java
```diff
@@ -0,0 +1,337 @@
...
+   * Check that fetchRelatedAdjectives() throws a RuntimeException
+   * when the JSON can't be parsed with the correct message signaling
+   * it's an issue with the JSON.
+   */
+  @Test 
+  public void jsonCannotBeParsed() throws Exception {
+    final String NOUN = "beach";
+    final int MAX_ADJECTIVES = 5;
+    final String QUERIES = "rel_jjb=" + NOUN + "&max=" + MAX_ADJECTIVES;
+    final String INPUT = "not JSON";
```
ykjchen, line 159: When I see "input" I think of an input to the client. IMO this would be easier if named RESPONSE, since this is the response of the HTTP request... to be even clearer, you could name this HTTP_RESPONSE_BODY.

backstory/src/test/java/com/google/sps/story/DatamuseRequestClientTest.java
```diff
@@ -0,0 +1,337 @@
...
+
+    // inject the url but with a response that's not proper JSON
+    injectURLConnectionMock(DEFAULT_URL, QUERIES, INPUT);
+
+    try {
+      String[] actualOutput = client.fetchRelatedAdjectives(NOUN, MAX_ADJECTIVES);
+    } catch (RuntimeException exception) {
+      final String EXPECTED_MESSAGE = "Could not parse the JSON received back from the Datamuse Query.";
+    
+      Assert.assertEquals(EXPECTED_MESSAGE, exception.getMessage());
```
ykjchen, line 171: I'd discourage asserting the exact value of a string. This type of thing is likely to change, and asserting against this could have an undesirable maintenance cost.

Also applies below.

backstory/src/test/java/com/google/sps/story/DatamuseRequestClientTest.java
```diff
@@ -0,0 +1,337 @@
...
+      Assert.assertEquals(JSONException.class, exception.getCause().getClass());
+      return; // this is the behavior we wanted to happen
+    }
+
+    Assert.fail("No RuntimeException was thrown when parsing incorrect JSON.");
+  }   
+
+  /** 
+   * Check that fetchRelatedAdjectives() works for when less
```
ykjchen, line 242: "works for" -> "succeeds"?

backstory/src/test/java/com/google/sps/story/DatamuseRequestClientTest.java
annemariecabs: That makes sense. For the sake of time, I'll keep this static, but this is good to keep in mind. Thanks!

backstory/src/test/java/com/google/sps/story/DatamuseRequestClientTest.java
```diff
@@ -0,0 +1,337 @@
...
+   * the input stream that will be returned from it will be
+   * the passed-in String response. Needed in order to test methods
+   * that use a URL. Code adapted from this tutorial: 
+   * https://claritysoftware.co.uk/mocking-javas-url-with-mockito/.
+   * 
+   * @param url the base url to inject canned results for
+   * @param queries the quer(ies) to send to that url
+   * @param response the response to return from that query
+   */
+  private void injectURLConnectionMock(String url, String queries, String response) {
```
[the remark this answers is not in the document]
annemariecabs: That makes sense. I will change this.

backstory/src/test/java/com/google/sps/story/DatamuseRequestClientTest.java
annemariecabs: I think I can now. I'll make the change.

backstory/src/test/java/com/google/sps/story/DatamuseRequestClientTest.java
annemariecabs: Ok that makes sense. I'll remove the error message validation.

backstory/src/test/java/com/google/sps/story/DatamuseRequestClientTest.java
annemariecabs: That sounds better. I'll make the change.

backstory/src/test/java/com/google/sps/story/DatamuseRequestClientTest.java
annemariecabs: That makes sense. I think I'll rename it to RESPONSE.

backstory/src/test/java/com/google/sps/story/DatamuseRequestClientTest.java
annemariecabs: Done

backstory/src/test/java/com/google/sps/story/DatamuseRequestClientTest.java
annemariecabs: Done

backstory/src/test/java/com/google/sps/story/DatamuseRequestClientTest.java
annemariecabs: Done

backstory/src/test/java/com/google/sps/story/DatamuseRequestClientTest.java
annemariecabs: Done

PR #166, opened 2020-08-18
Text Generation Backup Containers

UdayKalra:
Replaced single container URL with 5 URLs with a cycling method. Test also implemented to ensure no out of bounds exception in cycling. PR also includes a setPrefix() method to allow persistent use of an instance.


backstory/src/main/java/com/google/sps/story/StoryManagerImpl.java
```diff
@@ -58,6 +69,7 @@ public StoryManagerImpl(String prefix, int maxLength, Double temperature)
     this.maxTextLength = maxLength;
     this.temperature = temperature;
     requestFactory = new StoryManagerRequestFactoryImpl();
+    selectedUrlIndex = 0;
```
ykjchen, line 72: I don't think this is needed, since int fields are initialized to 0. If you want to make this explicit, you could include this above in the declaration:

```
private in selectedUrlIndex = 0;
```

backstory/src/test/java/com/google/sps/story/StoryManagerTest.java
```diff
@@ -85,6 +85,19 @@ public void invalidTemperature() {
     storyManager = new StoryManagerImpl(PREFIX_SAMPLE, SIZE_SAMPLE, 1000.0);
   }
 
+  @Test
+  public void cyclePastBounds() {
+    storyManager = new StoryManagerImpl(PREFIX_SAMPLE, SIZE_SAMPLE, TEMPERATURE_SAMPLE);
+    try{
+      for(int index = 0; index < 10; index++) {
```
ykjchen, line 92: The value of `10` here requires that this test case have internal knowledge of the number of URLs StoryManager may use. This would then require maintenance if the number of URLs changes. What would be a away to avoid that type of maintenance?

backstory/src/main/java/com/google/sps/story/StoryManager.java
```diff
@@ -30,4 +30,14 @@
...
+
+  /**
+   * Cycles selectedUrlIndex to a backup url.
+   */
+  public void cycleUrl();
+
+  /**
+   * Sets prompt for text generation.
+   */
+  public void setPrefix(String prefix);
```
ykjchen, line 42: This looks unused in this PR... what will it be used for?

backstory/src/main/java/com/google/sps/story/StoryManager.java
```diff
@@ -30,4 +30,14 @@
    * @param factory StoryManagerRequestFactory to use for HttpRequests.
    */
   public void setRequestFactory(StoryManagerRequestFactory factory);
+
+  /**
+   * Cycles selectedUrlIndex to a backup url.
+   */
+  public void cycleUrl();
```
ykjchen, line 37: Can you explain when this might be needed?

backstory/src/main/java/com/google/sps/story/StoryManagerImpl.java
```diff
@@ -144,4 +156,22 @@ private String makeRequestBody(String prefix, int maxLength, Double temperature)
     String convertedMap = gson.toJson(requestMap);
     return convertedMap;
   }
+
+  /**
+   * Cycles selectedUrlIndex to a backup url.
+   */
+  public void cycleUrl() {
+    if (selectedUrlIndex < 4) {
```
ykjchen, line 164: Using the literal 4 here makes this more likely to break if the serviceUrls array is updated. Could you get the size of the array instead?

backstory/src/main/java/com/google/sps/story/StoryManager.java
UdayKalra: If the container throws an error, the backstory servlet can switch to another container.

backstory/src/main/java/com/google/sps/story/StoryManager.java
UdayKalra: The backstory servlet will likely be re-using the storyManager object and cycling the URL, so this lets it change the prompt on each request without re-instantiating the object.

backstory/src/main/java/com/google/sps/story/StoryManagerImpl.java
UdayKalra: Done

backstory/src/test/java/com/google/sps/story/StoryManagerTest.java
UdayKalra: Changed to return boolean for indication of last URL.

backstory/src/main/java/com/google/sps/story/StoryManagerImpl.java
UdayKalra: Done

backstory/src/main/java/com/google/sps/story/StoryManager.java
```diff
@@ -30,4 +30,15 @@
    * @param factory StoryManagerRequestFactory to use for HttpRequests.
    */
   public void setRequestFactory(StoryManagerRequestFactory factory);
+
+  /**
+   * Cycles to next serviceUrl to a backup url.
```
annemariecabs, line 35: Nit: Should be newline in between description and tags.

backstory/src/main/java/com/google/sps/story/StoryManager.java
```diff
@@ -30,4 +30,15 @@
...
+   * @return False if called on last URL.
```
annemariecabs, line 36: Does this never return true? I think the purpose of this method could be a bit clearer if you specified when it returned true.

backstory/src/main/java/com/google/sps/story/StoryManager.java
annemariecabs, line 35: Is next serviceUrl the same as backup url? Slightly confused about this comment.

backstory/src/main/java/com/google/sps/story/StoryManagerImpl.java
```diff
@@ -45,6 +45,15 @@
   /** requestFactory - Builds and facilitates authenticated post requests. */
   private StoryManagerRequestFactory requestFactory;
 
+  /** serviceUrls - URLs for each story generation container */
+  private String[] serviceUrls = {"https://backstory-text-gen-1-pdaqhmzgva-uc.a.run.app",
```
annemariecabs, line 49: According to the Java style guide, I think that this array is supposed to be formatted slightly differently. E.g. the first String should not be on the same line as serviceUrls. However, if you just add a comma after the last String, then `make pretty` will format it for you the way the Java style guide does it.

backstory/src/main/java/com/google/sps/story/StoryManagerImpl.java
```diff
@@ -45,6 +45,15 @@
...
+      "https://backstory-text-gen-2-pdaqhmzgva-uc.a.run.app",
+      "https://backstory-text-gen-3-pdaqhmzgva-uc.a.run.app",
+      "https://backstory-text-gen-4-pdaqhmzgva-uc.a.run.app",
+      "https://backstory-text-gen-5-pdaqhmzgva-uc.a.run.app"};
+
+  private int selectedUrlIndex = 0;
```
annemariecabs, line 55: A comment here could be helpful (and technically it needs Javadocs even as a private field). It could be especially good to explain why you put 0 as its value (although I'm guessing that's just the default). Also, shouldn't setting it to 0 happen in the constructor, as this isn't a constant?

```suggestion
  /** the index of the current selected url in serviceUrls */
  private int selectedUrlIndex;
```

Then, instantiate it in the constructor.

backstory/src/main/java/com/google/sps/story/StoryManagerImpl.java
```diff
@@ -144,4 +153,28 @@ private String makeRequestBody(String prefix, int maxLength, Double temperature)
...
+    if (selectedUrlIndex < serviceUrls.length) {
+      selectedUrlIndex++;
+    } else {
+      selectedUrlIndex = 0;
+    }
+    return isLast;
+  }
+
+  /**
+   * Sets prompt for text generation.
```
annemariecabs, line 175: You should probably have a @param tag here.

```suggestion
   * Sets prompt for text generation.
   *
   * @param prefix the prefix to act as a prompt 
```

backstory/src/main/java/com/google/sps/story/StoryManagerRequestFactoryImpl.java
```diff
@@ -36,10 +36,9 @@
    *
    * @return HttpRequest Post Request
```
annemariecabs, line 37: Here, you need tags for your parameters and for the errors you're throwing. It's harder to understand these methods without these tags.

backstory/src/test/java/com/google/sps/story/StoryManagerRequestFactoryFakeImpl.java
```diff
@@ -46,7 +46,7 @@ public void setRequest(HttpRequest request) {
    *
    * @return HttpRequest Post Request
```
annemariecabs, line 47: As mentioned above, if you could just quickly add tags for these methods, it would make it easier to read and understand them.

backstory/src/main/java/com/google/sps/story/StoryManagerRequestFactory.java
```diff
@@ -36,5 +36,6 @@
    *
    * @return HttpRequest Post Request
    */
-  public HttpRequest newInstance(String requestBody) throws IllegalStateException, IOException;
```
annemariecabs, line 39: Sorry to repeat myself, but tags would be helpful here to.

backstory/src/test/java/com/google/sps/story/StoryManagerTest.java
```diff
@@ -85,6 +85,25 @@ public void invalidTemperature() {
     storyManager = new StoryManagerImpl(PREFIX_SAMPLE, SIZE_SAMPLE, 1000.0);
   }
 
+  @Test
+  public void cyclePastBounds() {
+    storyManager = new StoryManagerImpl(PREFIX_SAMPLE, SIZE_SAMPLE, TEMPERATURE_SAMPLE);
+    try{
```
annemariecabs, line 91: I think you're supposed to leave a space here.

```suggestion
    try {
```

backstory/src/test/java/com/google/sps/story/StoryManagerTest.java
```diff
@@ -85,6 +85,25 @@ public void invalidTemperature() {
...
+      boolean firstCycle = true;
+      
+      //Cycle through all URLs in first cycle.
+      while(firstCycle) {
```
annemariecabs, line 95: Similarly, according to Google style guide there should be a space after while (also one before cycle in the comments).

```suggestion
      while (firstCycle) {
```

backstory/src/test/java/com/google/sps/story/StoryManagerTest.java
```diff
@@ -85,6 +85,25 @@ public void invalidTemperature() {
...
+        firstCycle = storyManager.cycleUrl();
+      }
+
+      //Cycle once more.
```
annemariecabs, line 99: Once again, just adding the space.

```suggestion
      // Cycle once more.
```

backstory/src/main/java/com/google/sps/story/StoryManager.java
ykjchen: I see... I think making `prefix` mutable could introduce unexpected behavior, not with the current usage of this class, but consider this use case:

1. FooService is a user of StoryManager.
2. BarService is also a user of StoryManager.
3. BazService owns both FooService and BarService. When initializing FooService and BarService, BazService creates a new instance of StoryManager with prompt X, and uses that StoryManager to construct FooService and BarService.
4. Now FooService and BarService share the same StoryManager.
5. FooService decides at some point to change the prompt to Y.
6. BarService continues to use StoryManager assuming that the prompt is still X.

Perhaps a bit contrived, but I think it would be simpler if StoryManager was not mutable like this... it avoids the need to consider edge cases that may occur if the prompt is changed.

One solution may be to create a URLProvider:

```
class StoryManagerURLProvider {
  String getCurrentURL();
  void cycleURL();
}

class StoryManagerImpl {
  // The constructor accepts an instance of the URL Provider.
  StoryManagerImpl(StoryManagerURLProvider urlProvider, ...);

  boolean cycleURL() {
    urlProvider.cycleURL();
  }
}
```

Then you could have multiple StoryManager instances sharing the same URLProvider, and not need to worry about that type of mutation.

You would also probably want to make StoryManagerURLProvider thread-safe:
```
class StoryManagerURLProvider {
  synchronized String getCurrentURL();
  synchronized void cycleURL();
}
```

In case text generation calls become parallel in the future.

What do you think?

ykjchen: @UdayKalra ping

UdayKalra: Yeah, I agree. I'll set this up and PR it soon.

backstory/src/main/java/com/google/sps/story/StoryManagerURLProvider.java
```diff
@@ -0,0 +1,80 @@
...
+  };
+
+  /** Index of URL to provide. */
+  private int selectedURLIndex;
+
+  /**
+   * Initializes index for URL cycling.
+   */
+  public StoryManagerURLProvider() {
+    selectedURLIndex = 0;
```
ykjchen, line 50: I don't think this is needed, since `int` is by default initialized to 0?

backstory/src/main/java/com/google/sps/story/StoryManagerURLProvider.java
```diff
@@ -0,0 +1,80 @@
...
+  }
+
+  /**
+   * Cycles to next serviceUrl to an alternate container.
+   * Returns true when not last URL.
+   *
+   * @return False if called on last URL.
```
ykjchen, line 57: It looks like this method cycles whether or not the current URL is the last URL. Is there a need to expose to the client the fact that this cycled to the first URL? This "leaks" the implementation, in that now the client expects that there is some circular array. Ideally the client would have no knowledge of this, so that implementation can be updated without a breaking API change.

backstory/src/main/java/com/google/sps/story/StoryManagerURLProvider.java
```diff
@@ -0,0 +1,80 @@
...
+   */
+  synchronized public boolean cycleURL() {
+    boolean isLast = true;
+    if (selectedURLIndex == serviceURLs.length - 1) {
+      isLast = false;
+    }
```
ykjchen, line 63: The code above could be shortened to:
```
boolean isLast = selectedURLIndex != serviceURLs.length - 1;
```

backstory/src/main/java/com/google/sps/servlets/GPT2Servlet.java
```diff
@@ -0,0 +1,106 @@
...
+import javax.servlet.annotation.WebServlet;
+import javax.servlet.http.HttpServlet;
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+import org.json.JSONException;
+import org.json.JSONObject;
+
+/** Servlet that returns a generated story. */
+@WebServlet("/gpt2")
+public final class GPT2Servlet extends HttpServlet {
```
ykjchen, line 36: Does this belong in the demo folder?

backstory/src/main/java/com/google/sps/story/StoryManagerURLProvider.java
UdayKalra: Done

backstory/src/main/java/com/google/sps/story/StoryManagerURLProvider.java
UdayKalra: Done. Used getCurrentURL and tracked it for testing.

backstory/src/main/java/com/google/sps/story/StoryManagerURLProvider.java
UdayKalra: boolean removed all together.

backstory/src/main/java/com/google/sps/servlets/GPT2Servlet.java
UdayKalra: Modified wrong file. Attached correct one in new commit.

backstory/src/demo/java/com/google/sps/servlets/GPT2Servlet.java
```diff
@@ -38,6 +37,8 @@
   
   public static final int DEFAULT_MAX_STORY_LENGTH = 200;
   public static final Double DEFAULT_TEMPERATURE = 0.7;
+  private StoryManagerURLProvider URLProvider;
```
ykjchen, line 40: I think this field needs to be static and synchronized, since multiple servlet instances may be created. Maybe this would be a solution:
https://wiki.sei.cmu.edu/confluence/display/java/LCK05-J.+Synchronize+access+to+static+fields+that+can+be+modified+by+untrusted+code

backstory/src/demo/java/com/google/sps/servlets/GPT2Servlet.java
```diff
@@ -65,8 +66,12 @@ public void doPost(HttpServletRequest request, HttpServletResponse response) thr
     }
     String generatedText;
     try {
+      if (URLProvider == null) {
+        URLProvider = new StoryManagerURLProvider();
+      }
       StoryManager storyManager = new StoryManagerImpl(text, DEFAULT_MAX_STORY_LENGTH, DEFAULT_TEMPERATURE);
       generatedText = storyManager.generateText();
+      URLProvider.cycleURL();
```
ykjchen, line 74: Do you always want to cycle or just when there's an exception?

backstory/src/main/java/com/google/sps/story/StoryManagerImpl.java
```diff
@@ -41,6 +41,8 @@
   private int maxTextLength;
   /** The volatility of topics and style in generation. */
   private Double temperature;
+  /** Provider object of serviceURLs*/
```
ykjchen, line 44: Nit: space after URLs and before *

backstory/src/main/java/com/google/sps/story/StoryManagerURLProvider.java
ykjchen: Makes sense, I like your solution.

backstory/src/main/java/com/google/sps/story/StoryManagerURLProvider.java
```diff
@@ -0,0 +1,68 @@
...
+import com.google.auth.oauth2.IdTokenProvider;
+import java.io.IOException;
+
+/**
+ * Object for providing container URLs to StoryManager.
+ */
+public class StoryManagerURLProvider {
+    
+  /** serviceUrls - URLs for each story generation container */
+  private String[] serviceURLs = {
```
ykjchen, line 35: Can this be final and static?

backstory/src/test/java/com/google/sps/story/StoryManagerTest.java
```diff
@@ -144,4 +150,23 @@ public void correctRequestConfiguration() throws RuntimeException {
...
   }
+
+  @Test
+  /**
+   * Ensures Provider does not cycle past bounds.
+   */
+  public void cyclePastBounds() {
+    try {
+      String firstURL = URLProvider.getCurrentURL();
+      URLProvider.cycleURL();
```
ykjchen, line 161: I think you could use a do-while construct to avoid needing this statement twice:
https://docs.oracle.com/javase/tutorial/java/nutsandbolts/while.html

backstory/src/demo/java/com/google/sps/servlets/GPT2Servlet.java
UdayKalra: Done

backstory/src/demo/java/com/google/sps/servlets/GPT2Servlet.java
UdayKalra: On every occasion for the demo.

backstory/src/main/java/com/google/sps/story/StoryManagerImpl.java
UdayKalra: Done

backstory/src/main/java/com/google/sps/story/StoryManagerURLProvider.java
UdayKalra: Done

backstory/src/test/java/com/google/sps/story/StoryManagerTest.java
UdayKalra: Done

PR #176, opened 2020-08-19
8/19: Add image validation and fix issue where you can submit image before uploading image

### LINKED ISSUE #164: NullPointerException on hitting "Submit" with no uploaded image in Backstory alpha
Steps to reproduce:

1. Go to https://step174-step-2020.appspot.com/index.html
2. Do not upload an image.
3. Tap "Submit"

Result: You'll get an error page for a Java NullPointerException.

Expected: The Submit button should be disabled until an image is uploaded.

annemariecabs:
Add image validation and clean up minor front end critiques.

Fix #164 by not allowing users to submit the image upload form without submitting an image. First, does this by having the submit button be disabled and invisible until image is uploaded. Second, add validation that image is uploaded before sending the request (alerts user if image is not uploaded). Second method is unnecessary, but good extra security. Also, add validation that image submitted is a PNG or JPG/JPEG image, and alert user if not (also don't send post request/load loading graphic). Also add "here" to the image upload label to make UI more understandable.

### DIFF
--- a/backstory/src/main/webapp/script.js
+++ b/backstory/src/main/webapp/script.js
@@ -29,7 +29,7 @@
  * @return true, if image is valid, false otherwise
  */
 function checkForm() {
-  if (! validateImageUpload()) {
+  if (!validateImageUpload()) {
     return false;
   }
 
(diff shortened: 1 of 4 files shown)


backstory/src/main/webapp/script.js
```diff
@@ -153,3 +220,16 @@ function updateFileName() {
...
+
+/**
+ * Enables the submit button if there's a file in
+ * the image-upload field.
+ */
+function enableSubmitButton() {
+  const fileInput = document.getElementById('image-upload');
+  const submitButton = document.getElementById('submit-button');
+
+  if (fileInput.files.length > 0) {
```
aaronegolden, line 232: Could we use validateImageUpload here? That would get us the type checking of validateImageUpload and if the validation logic ever changes in the future it would keep the two pieces of code in sync.

annemariecabs: I can definitely add this. Thank you for the suggestion!

annemariecabs: Done

PR #190, opened 2020-08-27
8/26: Add image analysis to enable adding location to prompt

annemariecabs:
Add analysis of image with Landmark Detection feature of Cloud Vision. Adjust AnnotatedImage to store the labelAnnotations. Add testing for both of these adjustments. Resolves #118.

PR #191, opened 2020-08-27
8/27: Add Image Location into Prompt

### LINKED ISSUE #118: Add Image Geo-location Analytics

annemariecabs:
Update prompt manager so that it takes keywords and locations. Update tests accordingly. Add formatted location (location chosen is first location in list) to the prompt. Add tests accordingly. Integrates features built in #190 and helps resolve #118.

### DIFF
--- a/backstory/src/main/java/com/google/sps/story/PromptManager.java
+++ b/backstory/src/main/java/com/google/sps/story/PromptManager.java
@@ -13,6 +13,7 @@
 // limitations under the License.
 package com.google.sps.story;
 
+import com.google.common.collect.ImmutableList;
 import com.google.sps.story.data.*;
 import java.lang.IllegalArgumentException;
 import java.util.ArrayList;
@@ -26,9 +27,9 @@ import java.util.Scanner;
  */
 public final class PromptManager {
   /** Keywords for generation */
-  private final List<String> keywords;
+  private final ImmutableList<String> keywords;
   /** Locations for generation */
-  private final List<String> locations;
+  private final ImmutableList<String> locations;
   /** Word fetching/processing client */
   private PromptManagerAPIsClient wordAPIsClient;
   /** Randomness flag. */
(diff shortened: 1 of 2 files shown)


backstory/src/main/java/com/google/sps/story/PromptManager.java
```diff
@@ -71,7 +77,7 @@ public void isTemplateRandomized(boolean isTemplateRandomized) {
    */
   public String generatePrompt() {
     // Prepare story-like prefix.
-    String prompt = "Once upon a time, ";
+    String prompt = "Once upon a time" + getFormattedLocation() + ", ";
```
ykjchen, line 80: I feel like this method should be responsible for inserting spaces around the location text as needed. The `getFormattedLocation()` method shouldn't know how it's being used, and it not adding a prefixing space makes the method more flexible (you won't need to strip the space before using it in other places).

```
String locationString = getFormattedLocation();
if (locationString.length() > 0) {
  locationString = " " + locationString;
}
```

backstory/src/main/java/com/google/sps/story/PromptManager.java
```diff
@@ -36,14 +38,18 @@
...
+      throws IllegalArgumentException {
     // Check for null input.
-    if (keywords == null) {
-      throw new IllegalArgumentException("Input list cannot be null.");
+    if (keywords == null || locations == null) {
+      throw new IllegalArgumentException("Input lists cannot be null.");
     }
+
     this.keywords = keywords;
+    this.locations = locations;
```
ykjchen, line 52: Should probably make a (preferably immutable) copy, so that the values in `location` don't change while this instance is using them.

backstory/src/main/java/com/google/sps/story/PromptManager.java
annemariecabs: That makes sense. I'll make this change. Thanks!

backstory/src/main/java/com/google/sps/story/PromptManager.java
annemariecabs: That's a great suggestion. I'll add this.

backstory/src/main/java/com/google/sps/story/PromptManager.java
annemariecabs: Done

backstory/src/main/java/com/google/sps/story/PromptManager.java
annemariecabs: Done

